PR #300, opened 2014-12-20
Add a canonical URL to package pages.

chrisdotcode:
This is necessary because search engines will needlessly index older versions of
packages unless explicitly pointed to the newer version via a canonical link.


cartazio:
@chrisdotcode as @dcoutts  said, twiddle this so it build fine and it can be merged in!

chrisdotcode:
My apologies. I completely forgot about this. Thanks for the reminder; will finish within the week.

PR #356, opened 2015-05-07
Include readme

snoyberg:
All I've done is updated PR #333 to the latest master, and added an extra commit to keep both the description and the README content.

### DIFF
--- a/Distribution/Server/Features.hs
+++ b/Distribution/Server/Features.hs
@@ -165,11 +165,11 @@ initHackageFeatures env@ServerEnv{serverVerbosity = verbosity} = do
     packageContentsFeature <- mkPackageContentsFeature
                                 coreFeature
                                 tarIndexCacheFeature
+                                usersFeature
 
     packagesFeature <- mkRecentPackagesFeature
                          usersFeature
                          coreFeature
-                         packageContentsFeature
 
     userDetailsFeature <- mkUserDetailsFeature
                             usersFeature
@@ -250,7 +250,7 @@ initHackageFeatures env@ServerEnv{serverVerbosity = verbosity} = do
     htmlFeature     <- mkHtmlFeature
                          usersFeature
                          coreFeature
-                         packagesFeature
+                         packageContentsFeature
                          uploadFeature
                          candidatesFeature
                          versionsFeature
--- a/Distribution/Server/Features/Html.hs
+++ b/Distribution/Server/Features/Html.hs
@@ -9,7 +9,6 @@ import qualified Distribution.Server.Framework.ResponseContentTypes as Resource
 import Distribution.Server.Framework.Templating
 
 import Distribution.Server.Features.Core
-import Distribution.Server.Features.RecentPackages
 import Distribution.Server.Features.Upload
 import Distribution.Server.Features.BuildReports
 import Distribution.Server.Features.BuildReports.Render
@@ -20,6 +19,7 @@ import Distribution.Server.Features.Search
 import Distribution.Server.Features.Search as Search
 import Distribution.Server.Features.PreferredVersions
 -- [reverse index disabled] import Distribution.Server.Features.ReverseDependencies
+import Distribution.Server.Features.PackageContents (PackageContentsFeature(..))
 import Distribution.Server.Features.PackageList
 import Distribution.Server.Features.Tags
 import Distribution.Server.Features.Mirror
@@ -97,7 +97,7 @@ instance IsHackageFeature HtmlFeature where
 initHtmlFeature :: ServerEnv
                 -> IO (UserFeature
                     -> CoreFeature
-                    -> RecentPackagesFeature
+                    -> PackageContentsFeature
                     -> UploadFeature -> PackageCandidatesFeature
                     -> VersionsFeature
                     -- [reverse index disabled] -> ReverseFeature
@@ -176,7 +176,7 @@ initHtmlFeature ServerEnv{serverTemplatesDir, serverTemplatesMode,
 
 htmlFeature :: UserFeature
             -> CoreFeature
-            -> RecentPackagesFeature
+            -> PackageContentsFeature
             -> UploadFeature
             -> PackageCandidatesFeature
             -> VersionsFeature
@@ -199,7 +199,7 @@ htmlFeature :: UserFeature
 
 htmlFeature user
             core@CoreFeature{queryGetPackageIndex}
-            recent upload
+            packages upload
             candidates versions
             -- [reverse index disabled] ReverseFeature{..}
             tags download
@@ -243,7 +243,7 @@ htmlFeature user
                                       reportsCore
                                       download
                                       distros
-                                      recent
+                                      packages
                                       htmlTags
                                       htmlPreferred
                                       cachePackagesPage
@@ -433,7 +433,7 @@ mkHtmlCore :: HtmlUtilities
            -> ReportsFeature
            -> DownloadFeature
            -> DistroFeature
-           -> RecentPackagesFeature
+           -> PackageContentsFeature
            -> HtmlTags
            -> HtmlPreferred
            -> AsyncCache Response
@@ -455,7 +455,7 @@ mkHtmlCore HtmlUtilities{..}
            reportsFeature
            DownloadFeature{recentPackageDownloads,totalPackageDownloads}
            DistroFeature{queryPackageStatus}
-           RecentPackagesFeature{packageRender}
+           PackageContentsFeature{packageRender}
            HtmlTags{..}
            HtmlPreferred{..}
            cachePackagesPage
--- a/Distribution/Server/Features/PackageCandidates.hs
+++ b/Distribution/Server/Features/PackageCandidates.hs
@@ -22,6 +22,7 @@ import Distribution.Server.Features.TarIndexCache
 import Distribution.Server.Packages.Types
 import Distribution.Server.Packages.Render
 import Distribution.Server.Packages.ChangeLog
+import Distribution.Server.Packages.Readme
 import qualified Distribution.Server.Users.Types as Users
 import qualified Distribution.Server.Users.Group as Group
 import qualified Distribution.Server.Framework.BlobStorage as BlobStorage
@@ -30,7 +31,6 @@ import Distribution.Server.Packages.PackageIndex (PackageIndex)
 import qualified Distribution.Server.Framework.ResponseContentTypes as Resource
 
 import Distribution.Server.Util.ServeTarball (serveTarEntry, serveTarball)
-import qualified Data.TarIndex as TarIndex
 
 import Distribution.Text
 import Distribution.Package
@@ -39,9 +39,7 @@ import Data.Version
 import Data.Function (fix)
 import Data.List (find)
 import Data.Time.Clock (getCurrentTime)
-import Control.Monad.Error (ErrorT(..))
 import qualified Data.Vector as Vec
-
 
 data PackageCandidatesFeature = PackageCandidatesFeature {
     candidatesFeatureInterface :: HackageFeature,
@@ -159,7 +157,7 @@ candidatesFeature ServerEnv{serverBlobStore = store}
                              , updateAddPackage
                              }
                   UploadFeature{..}
-                  TarIndexCacheFeature{cachedPackageTarIndex}
+                  TarIndexCacheFeature{packageTarball, findToplevelFile}
                   candidatesState
   = PackageCandidatesFeature{..}
   where
@@ -373,11 +371,18 @@ candidatesFeature ServerEnv{serverBlobStore = store}
     candidateRender cand = do
            users  <- queryGetUserDb
            index  <- queryGetPackageIndex
-           mChangeLog <- packageChangeLog (candPkgInfo cand)
-           let showChangeLogLink = case mChangeLog of Right _ -> True ; _ -> False
-           render <- doPackageRender users (candPkgInfo cand) showChangeLogLink
+           let pkg = candPkgInfo cand
+           mChangeLog <- findToplevelFile pkg isChangeLogFile
+           mReadme    <- findToplevelFile pkg isReadmeFile
+           let changeLog = case mChangeLog of Right (_,_,_,fname,contents) -> Just (fname, contents)
+                                              _                            -> Nothing
+               readme    = case mReadme    of Right (_,_,_,fname,contents) -> Just (fname, contents)
+                                              _                            -> Nothing
+               render = doPackageRender users pkg
            return $ CandidateRender {
-             candPackageRender = render { rendPkgUri = rendPkgUri render ++ "/candidate" },
+             candPackageRender = render { rendPkgUri = rendPkgUri render ++ "/candidate"
+                                        , rendChangeLog = changeLog
+                                        , rendReadme = readme},
              renderWarnings = candWarnings cand,
              hasIndexedPackage = not . null $ PackageIndex.lookupPackageName index (packageName cand)
            }
@@ -415,13 +420,13 @@ candidatesFeature ServerEnv{serverBlobStore = store}
     serveChangeLog :: DynamicPath -> ServerPartE Response
     serveChangeLog dpath = do
       pkg        <- packageInPath dpath >>= lookupCandidateId
-      mChangeLog <- liftIO $ packageChangeLog (candPkgInfo pkg)
+      mChangeLog <- liftIO $ findToplevelFile (candPkgInfo pkg) isChangeLogFile
       case mChangeLog of
         Left err ->
           errNotFound "Changelog not found" [MText err]
-        Right (fp, etag, offset, name) -> do
+        Right (fp, etag, offset, name, _contents) -> do
           cacheControl [Public, maxAgeMinutes 5] etag
-          liftIO $ serveTarEntry fp offset name
+          liftIO $ serveTarEntry fp offset name -- TODO: We've already loaded the contents; refactor
 
     -- return: not-found error or tarball
     serveContents :: DynamicPath -> ServerPartE Response
@@ -435,20 +440,3 @@ candidatesFeature ServerEnv{serverBlobStore = store}
           serveTarball (display (packageId pkg) ++ " candidate source tarball")
                        ["index.html"] (display (packageId pkg)) fp index
                        [Public, maxAgeMinutes 5] etag
-
-    packageTarball :: PkgInfo -> IO (Either String (FilePath, ETag, TarIndex.TarIndex))
-    packageTarball pkginfo
-      | Just (pkgTarball, _uploadinfo) <- pkgLatestTarball pkginfo = do
-          let blobid = pkgTarballNoGz pkgTarball
-              fp     = BlobStorage.filepath store blobid
-              etag   = BlobStorage.blobETag blobid
-          index <- cachedPackageTarIndex pkgTarball
-          return $ Right (fp, etag, index)
-      | otherwise = return $ Left "No tarball found"
-
-    packageChangeLog :: PkgInfo -> IO (Either String (FilePath, ETag, TarIndex.TarEntryOffset, FilePath))
-    packageChangeLog pkgInfo = runErrorT $ do
-      (fp, etag, index) <- ErrorT $ packageTarball pkgInfo
-      (offset, fname)   <- ErrorT $ return . maybe (Left "No changelog found") Right
-                                  $ findChangeLog pkgInfo index
-      return (fp, etag, offset, fname)
--- a/Distribution/Server/Features/PackageContents.hs
+++ b/Distribution/Server/Features/PackageContents.hs
@@ -6,28 +6,29 @@ module Distribution.Server.Features.PackageContents (
   ) where
 
 import Distribution.Server.Framework
-import qualified Distribution.Server.Framework.BlobStorage as BlobStorage
 
 import Distribution.Server.Features.Core
 import Distribution.Server.Features.TarIndexCache
 
-import Distribution.Server.Packages.Types
 import Distribution.Server.Packages.ChangeLog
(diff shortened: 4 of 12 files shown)


bergmark:
:+1:

dcoutts:
Great, thanks for testing it.

hvr:
I'm eager to see this in action :)

dcoutts:
Ok, so I tried it out and poked about at various examples and unfortunately it's not going to be quite so simple. There's lots of packages with readmes that either duplicate or are just worse than their descriptions. So for the moment I've got it only linking to the readme, though rendering the markdown.

My suggestion for doing better is this:

 1. we need a way to limit the length of the description or readme (e.g. show first N "lines", link to the rest, or nice html5 expander) otherwise it pushed everything down the page.
 2. when the description is empty or ultra-short, use the readme inline, otherwise just link to it.

Anyway, all the code is merged, and the presentation is already hopefully an improvement. We can work on the visuals as a follow-up. Patches welcome.

snoyberg:
Well... that sounds like I just wasted my time, as did everyone else who worked on this PR. Can you let me know when this is deployed on the live site so I can see how useless this feature turned out to be?

dcoutts:
Oh I don't think it's a waste of time at all, it just needs more tweaking before we can use it for all packages. For some packages it of course looked great (those that have no description and a good readme, like many of your packages) and for other packages it looked terrible (those with both a description and a long or simialr readme).

It's deployed now.

dcoutts:
To be clear, I'm not saying what's deployed now is how things must remain. On the contrary. I wanted to get it merged, but looking at the results as it was it was a significant regression. So I think this half-way house is better than not applying it and asking people to revise and resubmit (which would be really momentum-sapping). This gives us a good position to make the extra tweaks so that we can use the readme inline in the places where it's appropriate and so that it looks good.

hvr:
Fwiw, we already had the pushing-down issue w/ http://hackage.haskell.org/package/lens which has quite a large description-field...

I think both, the show first N lines suggestion as well as the "when the description is empty or ultra-short, use the readme inline, otherwise just link to it." heuristic would be useful.

...and I'm still wondering how we can fix `cabal info` not showing any description anymore if ppl unaware of that start  omitting the `description` field in favour of a `README` file.

I assume we can now link to READMEs from the `description`-field on Hackage simply by relative hrefs i.e., `see <./readme README>` or something like that?

hvr:
@snoyberg btw, does the mardown renderer not support html tags in markdown, see e.g. the Examples section in http://hackage.haskell.org/package/lens-4.11/readme

hvr:
@snoyberg btw, here's an example of a rather inappropriate README for Hackage: http://hackage.haskell.org/package/filediff-1.0.0.2/readme

The problem I see is that some READMEs are more tailored towards use on GitHub, and contain information not relevant for display on Hackage. Maybe we should communicate the intent by having a `<!-- hackage description -->` header or something in the README, which would inform Hackage that the README content is meant to be displayed on Hackage.

snoyberg:
I didn't write the code to call the Markdown renderer, nor was I part of the decision to use that one. Switching to something better is trivial.

But `filediff` is a bad example. That README is an improvement: it has the same information, plus some extra information that doesn't harm anyone (and the Travis and code coverage information will in fact be very interesting to people). Remember: adding a README to a .cabal file is still a deliberate act.

I've already wasted more time on this PR and related issues than I care to. It's pretty obvious that decisions are being made after work is invested by others, without letting others be involved in that decision making process. "momentum-sapping" is certainly the right word here; letting the PR bitrot in the first place, or the lack of clear guidance to the people working on it in the first place, being prime examples. I'll go figure out others ways to be productive and improve Haskell documentation that don't involve this PR.

ekmett:
@snoyberg The main issue is that many packages out there have either a comprehensive description or a comprehensive `README`, or sometimes both. Prior to all of this talk of "momentum slapping" things seemed to be proceeding quite productively toward getting something in place.

e.g. `lens` has a 410 line `README` file on top of a considerable description. Rendering both in their entirety rather destroys the presentation of the package on hackage. Worse, the markdown parser in question doesn't handle half of the markup that is in the page.

Adding the `README` to the .cabal file _was_ a deliberate act. 2 years ago. Now that act would have retroactive consequences in terms of presentation. Most of my `README`'s for instance include things like travis build-status, contact info, etc. and were written for a different audience.

chrisdone:
To be clear, there might be a misunderstanding of intent here:

* Problem 1: We recognize that Haddock syntax is avoided by everyone and this makes people not bother to write helpful elaborations of their packages. Even right now you can hit a random package and find a URL that's mangled or speech marks or backslashes in a description field that makes it misleading. We are not dealing with an existing pristine documentation system.
* Problem 2: It's been like this for a *long time*.
* In contrast, people do write good Markdown README files, and on Github, where the README is rendered and dumped in full after the metadata, it looks great.
* Solution: do the same thing.
* Caveat: We *do* want the description field to go away and not be presented on the page.
* Caveat: We want to encourage people to write good READMEs, this means a *change* in how authors produce READMEs and are able to present their packages. That means considerations like "this README doesn't look that good or as good as the description field" is a trade-off you have to make to progress.
* It's also the case that READMEs are *already* written for different rendering scenarios, both for opening when you clone a project, and when you read on Github, and therefore why not also Hackage? I don't think anybody cares that much how much tailoring happens as long as the meat is there.
* I'd like to emphasize that linking to a README misses the fundamental point of this feature; we can already link to an external README and people often do; to their Github page. The point is to have Hackage's package page be the canonical place to get the package, see the metadata and read the documentation, so that authors can provide a good introductory experience to their users *thanks to* Hackage, not *despite* Hackage.

I think the problem of the length of the README can be avoided by putting it at the end of the page. The fact that some packages have massive metadata that you have to scroll down is an orthogonal problem with the package page itself that needs solving (e.g. tabs or w/e) elsewhen.

snoyberg:
@ekmett I'm not sure where things were moving along productively previously. I was involved from the beginning. This has been lethargic from the outset.

Regarding your comment about having README + description being too much and redundant information: I agree completely. I think it's a mistake. But the previous PR seemed to have stalled/bitrotted because of a single comment from Herbert about that. So I decided to try and overcome it by adding the description in this PR. I even commented in both the commit itself and this PR that I thought it was a problem.

> Most of my README's for instance include things like travis build-status, contact info, etc. and were written for a different audience.

I want to make sure I'm understanding your intent here correctly: we should make sure that the only way to get content onto a package's page is via a very difficult-to-use format (cabal+Haddock) and must be duplicated with the README content we all (yourself included) are using already, because there's extra information there that someone may not care to see on the Hackage page?

hvr:
Re `filediff`, I tend to agree that metrics such as code-coverage are interesting, I'm not convinced that the build-status of the current `master` tip is relevant for a versioned link (as the build-status does not necessarily refer to `filediff-1.0.0.2`), nor that the hackage-badge providing redundant information should be displayed on Hackage.

And fwiw, re "without letting others be involved in that decision making process", @dcoutts  did in fact reach out on `#hackage` a couple of hours ago and got ahold of @cartazio and @edwardk to consult regarding description/README

While adding READMEs via `extra-source-files` is a deliberate act, its effect is only now that it suddenly may be rendered on Hackage. I just believe that we may want to give users more control about how the description-field and the READMEs are to interact.

@chrisdone re

>  We do want the description field to go away and not be presented on the page.

I still think that the `description`-field serves are useful purpose (even though it could be merged with the synopsis field) as an abstract (think paper-abstract) for a package. I'd expect it to tell me the gist of a package w/o getting into too much detail. A one-line description (like the synopsis currently provides) is IMHO too short for that. Also the abstract should contain enough keywords to make it useful for a full-text search (which the synopsis is too terse for imho)

The problem w/ the description-only-in-README I see is that it's not part of the index-tarball anymore, hence you can't access it anymore via tooling such as `cabal info $PKG` (or think e.g. an in-Emacs package browser in the style of `M-x package-list-packages`), and that would force you to use a web-service or fetching some add-on meta-data for browsing/searching the package database offline...

snoyberg:
And just because this irks me so much: why are we having this conversation *now*? The issues about this have been open since September. Anyone could have commented about their concerns around implementing this back then. Instead, we're now- after three people spent significant time implementing the feature- having discussions that could have trivially been had before the work started. This is inconsiderate at best.

The result we have now- a hard-to-find link to a README- is something I have been able to implement myself for years by just putting a URL in my description field. There was no need to waste my time implementing it if we were just going to end up in this situation.

snoyberg:
@hvr The desire to have offline metadata about packages is one I've heard only two people express. Nonetheless, I already solved that problem for you: https://github.com/commercialhaskell/all-cabal-metadata. I'd send a PR to cabal about that, but I don't exactly have a great track record of writing PRs that actually get included in a useful way.

ekmett:
@snoyberg The only reason I got involved in this discussion at all was because I happened to look at the #hackage channel when @dcoutts was discussing things with @cartazio about how it looked in practice.

I think we all like the general idea, but there is a great deal of concern about precisely the right way to go about this.

Automatically picking between the description and the `README` is awkward and ad hoc. As is picking just a prefix of both with a link to the full thing. Including all of both can often lead to too much wording. Other times they are written for different audiences.

Putting up a link to the `README` while looking for a better way to present things is a small step but gets the code integrated and tested and at least at first does no harm.

It leaves open options such as

* including the first few lines of each like mentioned above or of one or the other using ad hoc means
* specifying version ranges for which we want to show one or the other or both in the hackage meta-data 
* some horrible option like "adding a field to the .cabal file to say what you want to show so that in 5 years it can all work".
* saying it applies to new packages going forward, but that we won't change the front pages of old packages

All of those options kind of suck, but you have to admit that the proposal as it stands has rough edges of its own.

Why are these coming up now rather than back when this was first proposed? It isn't an attempt to curb stomp this idea or block momentum. It is that folks finally had a concrete proposal to poke holes in. They could see what it looked like and some of the results gave people pause.

mpickering:
FWIW, it was a conscious decision to disallow raw HTML, it can be changed by a single flag.

What you describe as "Hackage's inability to display documentation correctly" is actually a result of the incredible fragmentation of markdown specifications. [Pandoc](http://hackage.haskell.org/package/pandoc-1.13.1/readme) is another example of a package which looks terrible with these changes. The README is written in it's own markdown dialect.

wiz:
> What you describe as "Hackage's inability to display documentation correctly" is actually a result of the incredible fragmentation of markdown specifications.

Wouldn't it be great If we could have a common markdown [specs](http://commonmark.org/) and make a haskell [converter](http://hackage.haskell.org/package/cmark) for them?

dcoutts:
>  It's pretty obvious that decisions are being made after work is invested by others, without letting others be involved in that decision making process. "momentum-sapping" is certainly the right word here; letting the PR bitrot in the first place, or the lack of clear guidance to the people working on it in the first place, being prime examples. I'll go figure out others ways to be productive and improve Haskell documentation that don't involve this PR.

@snoyberg Look, this was certainly not my intention, rather the opposite. Yes this patch had been hanging around for some time (partly because it was quite large and partly because we didn't have good answers to this question of when to use the README). Myself and others greatly appreciate that you and @mpickering rescued @conklech's patch from bitrot. I spent some hours reviewing and then testing the patch, and then looking at this policy question of when to show the README. My priority was to get the patch in so we didn't suffer further bitrot and disheartenment for the patch authors. The detail of when the README is shown did not need o hold up getting the bulk of the code merged.

So yes, I could have asked you and @mpickering and people to go back and think about this before merging. But I thought it was better to ask that after merging. That's what I did. I did also ask other people for an immediate opinion, that wouldn't delay getting the code merged. That's why @ekmett and @cartazio gave me their 2c, because I asked for immediate feedback on IRC.

> without letting others be involved in that decision making process

That's what we're doing right now, after having merged the bulk of the code, leaving just the policy question to be decided. I'm sorry that I apparently gave the opposite impression.

As for why now, rather than right at the beginning. Honestly, it wasn't until I looked at the result that it was obvious that using both the description and the readme together gave bad results in many cases. Could this have been predicted from the start? I guess.

But seriously it's not that huge an issue. Deciding when to show the README is not a massive change to the code here, and not a waste of everyone's work. I've made a suggestion above about a policy of when to show the README inline, and that would not be hard to do. The issue of long READMEs/descriptions was in fact identified much earlier (when @mpickering picked up the patch).

dcoutts:
@chrisdone Thanks for the clarifications. Yes, I concur with your identification of Problem 1, 2 and the solution, and I have no problem with the ultimate goal of letting people use a README in place of the description from the .cabal file.

> * Caveat: We do want the description field to go away and not be presented on the page.
> * Caveat: We want to encourage people to write good READMEs, this means a change in how authors produce READMEs and are able to present their packages. That means considerations like "this README doesn't look that good or as good as the description field" is a trade-off you have to make to progress.

So I don't think there's a misunderstanding of the goals here. The intermediate problem is that while many packages do have good READMEs (and some poor descriptions), many packages do not have good READMEs. We can encourage a move to using README, but if we were to flip the switch on all old packages the results are pretty horrible (I've seen them). Hence my suggestion that we use the README when the description is empty/ultra-short. That will work now for many packages, and makes it easy to opt-in, even for released packages (by editing the .cabal file). To further encourage the switch we could have Cabal add a "description-file: README" field, and tell authors they should either have a `description` or a `description-file` (and obviously automatic support to add it to the tarball etc).

> * I'd like to emphasize that linking to a README misses the fundamental point of this feature; we can already link to an external README and people often do; to their Github page. The point is to have Hackage's package page be the canonical place to get the package, see the metadata and read the documentation, so that authors can provide a good introductory experience to their users thanks to Hackage, not despite Hackage.

As I've tried to make clear, the current state is just a half way house, to get the code merged, pending working out the detail of when to show the README inline.

Perhaps it'd have been less controversial if I'd merged the code but simply disabled the new feature initially and not added the README link. I thought it was helpful but apparently it has confused matters about what the ultimate intention is. Sorry about that.

And yes, I agree with you about the length issue. A page layout redesign would make it a lot easier to deal with long descriptions and READMEs. The current layout is predicated on a modest description. A layout that someone mocked up before http://code.haskell.org/~duncan/hackage-server-new-chrome/package.html has the properties at the side which leaves a lot more room for the description/readme. I think something like this would be nice. Patches welcome :-)

dcoutts:
@wiz We're spoilt for choice! :-) We've got pandoc, cheapskate and cmark. No problems here with switching to a different one if it gets us generally better results. That part of the code is quite well isolated, so feel free to send a patch.

wiz:
> A layout that someone mocked up before http://code.haskell.org/~duncan/hackage-server-new-chrome/package.html has the properties at the side which leaves a lot more room for the description/readme. I think something like this would be nice. Patches welcome :-)

Oh, this is nice. I would like to see a modern CSS framework (BS3, materialize, etc...) being utilized. Is there a repo i can fork?

> We're spoilt for choice! :-) We've got pandoc, cheapskate and cmark.

I think we should aim for `cmark` since it uses a reference implementation for a long-term standard and not yet another attempt at "getting markdown right", main culprit being there's no "right" for markdown.

dcoutts:
@wiz I'd love someone to pick up this redesign work. There's no repo, it was done ad-hoc by someone. I just preserved it. So look at the files in http://code.haskell.org/~duncan/hackage-server-new-chrome/

> I think we should aim for cmark ...

Vote noted. I don't have a particular opinion on this, whatever works. Hopefully we can decide quickly and just do it. Shouldn't be hard to change.

In fact perhaps while we've got the README not yet being on the front page we should take the opportunity to try the other markdown impls and then pick the one that works best for the files we've got. Pull reqs accepted!

jgm:
My 2 cents on pandoc vs. cheapskate vs. cmark (since I'm the author of all of them):

- If you want to support a wide range of features (e.g. various table formats, math, footnotes, definition lists), then **pandoc** is currently the best choice.  Cons:  it's a big dependency, it's not super-fast, and the Markdown parser doesn't have great performance characteristics (it can be gummed up by some [pathological inputs](https://github.com/jgm/pandoc/issues/1735)).  If you use it, you might want to put it under a timeout to ensure that people can't tie up the system with malicious inputs.

- If it's okay to have core Markdown features (plus fenced code blocks), then I'd recommend **cmark**. It's very well tested (including extensive fuzz testing, a week-long run with AFL).  It implements a well defined and well thought-out version of Markdown, which has already been implemented in a dozen languages and which will probably eventually become the core of github's Markdown.  The parser is extremely fast and efficient. The library can produce either HTML or a Haskell structure that can be manipulated and rendered as you wish. Cons:  it's not pure Haskell, but wraps a C library.

- If you want something lightweight, and it has to be pure Haskell, then maybe **cheapskate** is a reasonable choice.  But cheapskate's parser isn't as good as cmark's -- it's slower, and it can be gummed up by some pathological inputs.  It's also much less well tested, and it gives results I'd consider "wrong" on some inputs.  Now that we have cmark, maintaining cheapskate is fairly low on my priority list.

Note that if you use pandoc or cmark, you should combine it with xss-sanitize.

You might also consider the possibility of rendering client-side using jgm/commonmark.js, which should render exactly the same way `cmark` does and is faster than pandoc.

dcoutts:
@jgm thanks very much for the detailed advice.

snoyberg:
I discussed this feature at length with Duncan just now, and he asked me to write in my thoughts. I have exactly one thing I care about, which is getting the README content onto the Hackage package pages, so that I can start encouraging package authors to easily improve the documentation for their packages. I'm OK if the content appears below the metadata. I'm OK if there's an expander added. In fact, I haven't seen a single suggestion here that I'm opposed to. I just want something to happen.

dcoutts:
So here's my current suggestion:

 * keep displaying the description (though we may still want to limit it's length)
 * display the readme inline below the properties (before or after the module list?)

I think if we have both the description and reame, we can encourage people to have short descriptions, and put all the longer detail into the readme/extended docs. The original idea of the description was never to be a tutorial or whatever but to be a simple paragraph summary (with the synopsis being a very terse one-line summary). Since we've always only had the description on hackage and not had a good place for other docs (haddock api docs arn't great for intro tutorials and other misc stuff), then this has encouraged people to stuff more and more things into the description, to the point where they suffer from the poor markup and we have duplication between description and reame (and consequent annoyance from people about that repetition, and the poor markup etc). On the other hand people have pointed out that the description does serve a purpose, and has a slightly different audience and context. So hopefully if we do display both we can get the descriptions back down to a managable size (e.g. one paragraph) and give lots of space and nicer markup for more and better intro & misc docs.

If people think this is a good approach (short description, everything else in readme) then we can modify Cabal's Q/A checks to encourage people in that direction, e.g. complaining about long descriptions and pointing people towards using markdown readmes.

Also this approach has the advantage that it's easy to implement, which is important to get it done swiftly.

Thoughts?

chrisdone:
Agreed with all that. I like this approach. I'd suggest after the module list and bite the bullet on packages like lens where the module list is super long.

snoyberg:
Another "like Github" thing we could do: add a "Read more" link right after the synopsis or description that would take you down to the README content.

chrisdone:
Here is what the lens package looks like with the README taken from Github and dumped at the end of the page before the footer:

lens_lenses _folds_and_traversals_hackage_-_2015-05-19_18 00 57

dcoutts:
> Another "like Github" thing we could do: add a "Read more" link right after the synopsis or description that would take you down to the README content.

@snoyberg I like that. I was also a little worried that if we put it at the end that people simply will not notice. So I could replace the "readme" link to the out-of-line reamde with a "read more" link to that anchor within the page.

@chrisdone ok, thanks for the vote. Yeah, lens is crazy, but once we've got the readme in, I'm sure Edward can be persuaded to prune the description a bit (aeson is another offender in this area).

chrisdone:
@dcoutts By including the screenshot I meant more in a "this is okay" way. It's not that bad. So I think this'll work out nicely.

dcoutts:
@chrisdone yes, I misread at first. I get what you mean. And the good thing is that we can edit descriptions in the .cabal files to prune them where appropriate.

ekmett:
I'm willing to take a hacksaw to the descriptions in my `.cabal` files.

ekmett:
Does the ability to edit package meta-data in a revision extend to editing the descriptions or just the dependency stuff? If not could it be extended there?

dcoutts:
@ekmett I believe I set it up so that editing the descriptions and (other similar metadata) is allowed.

The principle being we're fairly liberal about changes to data that doesn't affect the build, and very cautious on stuff that does affect the build (only deps I think, nothing else iirc).

mboes:
> So here's my current suggestion:
> 
> * keep displaying the description (though we may still want to limit it's length)
> * display the readme inline below the properties [after the module list]
>
> [...] If people think this is a good approach (short description, everything else in readme) then we can modify Cabal's Q/A checks to encourage people in that direction, e.g. complaining about long descriptions and pointing people towards using markdown readmes.
>
> Also this approach has the advantage that it's easy to implement, which is important to get it done swiftly.

That means two fronts:

* Hackage changes, which given lack of any opposition could go live right now.
* Cabal changes, meant to goad people towards descriptions and README's that play nice with the above layout choice.

@dcoutts What's the status on either of those fronts? Need any help on either?

dcoutts:
@mboes the change to put the README inline is now done (pending deployment).

Changes to cabal to encourage good synopsis & description is yet to do. A pull req on that would be appreciated.

I think the rough consensus on the latter is this:

 * one-line synopsis
 * description should be a paragraph or two
 * longer intros etc in the README

And in particular, changelogs should not go in the description, they should go in a changelog(.md). A message in cabal check could perhaps also mention that markdown is encouraged for changelogs and readmes. The current hackage code will render markdown files as markdown, and all other extensions as plain text.

Rationale: synopsis is useful in many places for a very brief idea of what a package is about. Description is again useful for users to quickly get an idea about whether the package is relevant to them. The description gets downloaded to all clients so can be used in list/search features, which is harder for READMEs that live inside package tarballs.

PR #373, opened 2015-06-05
Fix preferred versions colors

potomak:
Colors of version links at /package/:package/preferred should follow the rule:

> Blue versions are normal versions. Green are those out of any preferred version ranges. Gray are deprecated.

### DIFF
--- a/Distribution/Server/Pages/Package.hs
+++ b/Distribution/Server/Pages/Package.hs
@@ -279,7 +279,7 @@ renderVersion (PackageIdentifier pname pversion) allVersions info =
                                ++ map versionedLink laterVersions
         versionedLink (v, s) = anchor ! (status s ++ [href $ packageURL $ PackageIdentifier pname v]) << display v
         status st = case st of
-            NormalVersion -> []
+            NormalVersion -> [theclass "normal"]
             DeprecatedVersion  -> [theclass "deprecated"]
             UnpreferredVersion -> [theclass "unpreferred"]
         infoHtml = case info of Nothing -> noHtml; Just str -> " (" +++ (anchor ! [href str] << "info") +++ ")"
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -41,6 +41,15 @@ a[href]:link { color: rgb(196,69,29); }
 a[href]:visited { color: rgb(171,105,84); }
 a[href]:hover { text-decoration:underline; }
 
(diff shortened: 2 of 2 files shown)


lfairy:
Can you paste a screenshot? :wink:

By the way, have you taken color-blind users into account? While I do support the gist of this change, I'm not sure if this implementation will be accessible. The [Wikipedia Manual of Style][1] recommends against using color alone for this reason.

I suggest adding some secondary formatting e.g. **bold** for preferred and ~~strike-through~~ for deprecated packages.

[1]: https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Accessibility#Color

hvr:
@lfairy making preferred bold would be too noisy, conflicts with the "current version" highlighting, and doesn't match the data-model imho; if you look at the classes, there's no "preferred" css class, only "normal", "unpreferred" and "deprecated"; so while strikethrough seems good for "deprecated", we need to come up with some lightweight style for "unpreferred"

potomak:
@lfairy here are two screenshots:

Package

Preferred versions

Well, I didn't take color-blind users into account, can you help me with that? I just followed the coloring suggested by the paragraph by making permutations of RGB of original link colors.

lfairy:
@potomak Cool, thanks.

Not sure about the choice of green for non-preferred, since green seems like a "safer" color than blue. Though I'd like a second opinion on that.

I think making non-preferred versions *italic* and deprecated versions ~~strike-through~~ would work. That would address @hvr's comment (hopefully italics are lightweight enough).

Alternatively you can make the non-preferred versions just grayed out, which might be less intrusive than italics. Then the deprecated versions can be gray *and* crossed out, so it's like they're super-non-preferred. If there's enough contrast between those and the normal versions, then it could work.

As for accessibility: the trick is to add some indicator other than hue/saturation. Changing the font style/weight is one way; making the text lighter or darker is another. I'm sure there are other ways; what I wrote above is just what came to mind first.

potomak:
@lfairy maybe we could ask @gracenotes for suggestions, I think he's the first author of this feature, see https://github.com/haskell/hackage-server/commit/7748340236bc9c744d901652dc84143d133e8095.

dcoutts:
I think it doesn't really matter what the original comment said about the colours. What's more important is picking something reasonable. My own 2p is: not too garish :-)

gbaz:
I think this was superceded by the other merged PR and so I'm closing?

PR #394, opened 2015-07-20
Add "Manual" column to the package flag table

grayjay:
Here is a screenshot from aeson:

aeson

This column might need an explanation.  Would it be useful to mention manual flags in the section of the Cabal User Guide that "More info" links to?

http://www.haskell.org/cabal/users-guide/installing-packages.html#controlling-flag-assignments

Manual flags are currently only described in the page on developing packages:

https://www.haskell.org/cabal/users-guide/developing-packages.html#configuration-flags

### DIFF
--- a/Distribution/Server/Pages/Package.hs
+++ b/Distribution/Server/Pages/Package.hs
@@ -225,12 +225,14 @@ renderPackageFlags render =
           ,tbody << map flagRow flags]
         flagsHeadings = [th << "Name"
                         ,th << "Description"
-                        ,th << "Default"]
+                        ,th << "Default"
+                        ,th << "Manual"]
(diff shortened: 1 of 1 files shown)


hvr:
I'd suggest a variation (which departs from the config-file syntax):

Name the column something like "Type", and use the label "manual" and "automatic" instead of True/False respectively.

grayjay:
Here is an updated screenshot.  I like "Manual" and "Automatic", but I was hoping to come up with a more descriptive heading.  Does anyone have any suggestions?

aeson-2

bergmark:
I wouldn't worry that much about the header name, the values makes it clear.

dcoutts:
Thanks!

PR #428, opened 2015-10-03
fix mirror client build failure

bennofs:
hackage-security changed it's API, which broke the mirror client build. This
PR fixes that.


kosmikus:
Thanks.

PR #435, opened 2015-10-12
Added stack.yaml files freezing the dependencies

berdario:
one for GHC 7.8, symlinked as the default and one for GHC 7.10

Maybe this is useful to others as well?

Upon suggestion from Duncan, I also prepared a cabal.config that contains the list of frozen dependencies, but... well, I created 2:

- [one from plain `cabal freeze`](https://gist.github.com/berdario/eaaf96f69accbb15d34e)
- [this one](https://gist.github.com/berdario/e3053e90877c2c99efa2) from `stack list-dependencies`, and by manually adding the versions for packages that come with ghc (is that a good idea?) The end result is that a version is different (ed25519's), rts ==1.0 is missing, and a bunch of dependencies needed for testing have been added

Not sure which of the 2 would be the desired one.

### DIFF
--- a/stack-7.10.yaml
+++ b/stack-7.10.yaml
@@ -1,0 +1,152 @@
+# For more information, see: https://github.com/commercialhaskell/stack/blob/release/doc/yaml_configuration.md
+
+# Specifies the GHC version and set of packages available (e.g., lts-3.5, nightly-2015-09-21, ghc-7.10.2)
+resolver: ghc-7.10.2
+
+# Local packages, usually specified by relative directory name
+packages:
+- '.'
+
+# Packages to be pulled from upstream that are not in the resolver (e.g., acme-missiles-0.3)
+extra-deps:
+- HStringTemplate-0.8.3
+- HTTP-4000.2.20
+- HUnit-1.3.0.0
+- HaXml-1.25.3
+- QuickCheck-2.8.1
+- SHA-1.6.4.2
+- acid-state-0.13.0
+- aeson-0.10.0.0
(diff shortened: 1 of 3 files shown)


berdario:
Whoops, I just realized now that a similar PR already exists:

https://github.com/haskell/hackage-server/pull/432

still, this PR/discussion might be useful for getting a stack.yaml for ghc-7.8 (does anyone want that?)
Or for picking up one of the 2 `cabal.config`s (again, not sure if people really need it)

zudov:
In my PR, I've just put pretty much what `cabal init && cabal solver --write-stack-yaml` gave me, and it worked fine.

Do you think it's really necessary to put so many `extra-deps`?

I think it would be much better to use packages from lts snapshots as much as possible, and put a `cabal.config` which would freeze cabal to the same versions as in the snapshot.

zudov:
Oh, sorry, I didn't notice that you are using ghc-resolver, that explains the amount of extra-deps.

Then why not to use `lts` resolver?

berdario:
Uh, nice. I didn't know of the existence of `cabal solver --write-stack-yaml`

I just went this way after looking at the [ `stack.yaml` for cabal](https://github.com/haskell/cabal/commit/fcdfb308ee60c4971d64b9aad954a6044e6a1820), but

> I went the direction of no curation and explicit packages in the
extra-deps, as I thought that would make more sense for a project low on
the dependency chain

Might not apply for hackage-server.
Otherwise, I'd prefer the `lts` resolver as well

PR #459, opened 2015-12-25
Fix the build

edsko:
PR #454 broke the build; compile error:

```
Distribution/Server/Features/Search/SearchEngine.hs:364:7:
    Couldn't match expected type ‘(Maybe key,
                                   [(BM25F.Explanation field feature Term, key)])’
                with actual type ‘[(BM25F.Explanation field feature Term, key)]’
```

(both on the `master` branch here and on the original `exact-match` branch from that PR). This PR fixes the compile error, but since I am not the author of the original PR this is a bit of a guess.

The crucial part of the fix is at https://github.com/haskell/hackage-server/pull/459/files#diff-8bb91ba499325cf56b275cbaed19e47dR362 .

Pinging @mpickering .


mpickering:
Thanks Edsko, I just found these changes uncommitted in my local copy. Sorry for breaking this!

edsko:
This fails for the standard reason (security tests). This commit will be part of #412 .

PR #462, opened 2016-01-04
add some context as to what ICU is

bollu:
I didn't know what `ICU` was, so I spent a minute updating the docs so they explain what the library install does and why `text-icu` needs them

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ This is the `hackage-server` code. This is what powers <http://hackage.haskell.o
 ## Installing ICU
 
 ICU stands for "International Components for Unicode". The `icu4c` is a set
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,6 +6,11 @@ This is the `hackage-server` code. This is what powers <http://hackage.haskell.o
 
 ## Installing ICU
 
+ICU stands for "International Components for Unicode". The `icu4c` is a set
+of libraries that provide Uncicode and Globalization support.
```
zudov: "Uncicode"

bollu: ouch, should have proofread it. Will change

bollu:
I was wondering if this could be merged. The build is failing for unrelated reasons AFAICT

PR #492, opened 2016-04-28
compile failure workaround

hvr:


hvr:
I believe this solves the same issue as #486 but differently...

PR #499, opened 2016-05-26
Add test-suite and benchmark tags

### LINKED ISSUE #498: Add "test-suite" and "benchmark" tag
Currently, Hackage automatically tags all packages with `library` that have library stanzas in their `.cabal` file, and similarly, it tags packages that have executables with `program`. Could something likewise be done with `test-suite` and `benchmark` tags? In particular, it would be super-useful for me to be able to tell from a glance all of the Hackage packages that have a benchmark (and would make it easier for me to grab _only_ those packages to obtain benchmark numbers).

RyanGlScott:
Implementing this mostly involves plumbing. The logic for determining whether to print "library", "library and program", "library, programs, tests and benchmarks", etc. became more complicated, so I factored it out into `Distribution.Server.Pages.Util`.

Fixes #498.

### DIFF
--- a/Distribution/Server/Features/Html.hs
+++ b/Distribution/Server/Features/Html.hs
@@ -1614,7 +1614,7 @@ mkHtmlSearch HtmlUtilities{..}
             renderSearchResult :: PackageItem -> Html
             renderSearchResult item = li ! classes <<
               [ packageNameLink pkgname
-              , toHtml $ " " ++ ptype (itemHasLibrary item) (itemNumExecutables item)
+              , toHtml $ " " ++ ptype
               , br
               , toHtml (itemDesc item)
               , br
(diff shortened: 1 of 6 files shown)


hvr:
@dcoutts @RyanGlScott merging this broke the build:

```
[129 of 147] Compiling Distribution.Server.Features.PackageList ( Distribution/Server/Features/PackageList.hs, dist/build/hackage-server/hackage-server-tmp/Distribution/Server/Features/PackageList.o )

Distribution/Server/Features/PackageList.hs:79:14:
    Constructor â€˜PackageItemâ€™ should have 9 arguments, but has been given 7
    In the pattern: PackageItem a b c d e f g
    In an equation for â€˜memSizeâ€™:
        memSize (PackageItem a b c d e f g) = memSize7 a b c d e f g
    In the instance declaration for â€˜MemSize PackageItemâ€™

Distribution/Server/Features/PackageList.hs:83:24:
    Couldn't match expected type â€˜PackageItemâ€™
                with actual type â€˜Int -> Int -> PackageItemâ€™
    Probable cause: â€˜PackageItemâ€™ is applied to too few arguments
    In the expression: PackageItem pkg Set.empty Nothing "" 0 False 0
    In an equation for â€˜emptyPackageItemâ€™:
        emptyPackageItem pkg
          = PackageItem pkg Set.empty Nothing "" 0 False 0
```

RyanGlScott:
Odd, I could have sworn it built locally successfully... in any case, I've opened #505 to atone.

PR #506, opened 2016-06-09
Use cryptohash-sha256 for SHA256

hvr:
This avoids the hard to constrain cryptohash/cryptonite/byteable/memory
dependency which is a bit of a moving target and uses the simpler and
focused `cryptohash-sha256` package instead. This is also what
hackage-security and therefore cabal-install uses.


hvr:
@dcoutts don't merge yet; I'd like to verify something first

hvr:
@dcoutts The new `HashTests` testsuite still passes, so I'm confident enough that this refactoring has no observable changes and therefore is binary compatible. So I'm going to merge this now.

PR #516, opened 2016-06-30
Prevent package README images from overflowing

yamadapc:
This limits package README `img` tags to 100% width, which prevents them
from overflowing. A screenshot:


hvr:
...and what would the screenshot look after the patch is applied? :-)

yamadapc:
stack-run__an_equivalent_to_cabal_run_for_stack__and_1__tmux

PR #532, opened 2016-08-31
Cache new documentation for short periods

treeowl:
Previously, the cache time for all documentation was set to
one day. Cache it, instead, for as long as it's been since it
was last modified, but no less than ten minutes and no more than
a day. This takes into account the fact that documentation is
particularly likely to be modified soon after it was uploaded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,6 @@ dist
 tags
 *.swp
 .cabal-sandbox
+.stack-work
 cabal.sandbox.config
 TAGS
--- a/Distribution/Server/Features/Documentation.hs
+++ b/Distribution/Server/Features/Documentation.hs
@@ -32,6 +32,9 @@ import Data.Function (fix)
 
 import Data.Aeson (toJSON)
 
+import Data.Time.Clock (NominalDiffTime, diffUTCTime, getCurrentTime)
+import System.Directory (getModificationTime)
+
 -- TODO:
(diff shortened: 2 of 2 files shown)


treeowl:
Quick question: does the cache time grow too quickly? Should I use one-half or one-third the age instead?

dcoutts:
> Quick question: does the cache time grow too quickly? Should I use one-half or one-third the age instead?

I don't have a strong opinion, you're welcome to adjust the scaling if you like, otherwise I'm happy to merge as is.

treeowl:
I think I'll slow it down a bit. Give people a few days before changes get
too painful. I'd love to use some pretty mathematical function that would
give it a nice sensible shape. Any thoughts?

dcoutts:
@treeowl could use a shallow exponential (with a min and max as you do now).

treeowl:
Yeah, that seems okay. I'll set one up.

treeowl:
There you go. I couldn't find any remotely pretty way to calculate the function because `NominalDiffTime` isn't `RealFloat`. But if my calculations are correct (and they seem to be), this should give a smooth progression from ten minutes to a maximum of one day (reached after four days).

treeowl:
@dcoutts Is this ready to merge?

dcoutts:
@treeowl thanks! sorry was away at a hackathon.

treeowl:
Hope you had fun!

PR #552, opened 2016-10-23
updates to central index page

gbaz:
This is intended to be deployed with the Summer of Haskell changes, as it points to the new tabled browsing page, puts a new link to the old page, deletes links to some pages superseded by the new tabled browsing interface, and includes a few other changes in passing to update a few links, and add a little more info.


gbaz:
While I think the first set of changes should be uncontroversial, I also included another patch, https://github.com/haskell/hackage-server/pull/552/commits/f3e404eda30bb995ef1e39890c3762737fced139 which I think streamlines the bottom "Hackage 2" section without losing anything important. I hope what I did is modest enough that its uncontroversial as well, but a bit more attention may want to be paid there as well :-)

gbaz:
@hvr can you push this with the next hackage deploy? This moves the /browse UI to the first link and puts the classic categories view in the other section, among many other changes...

PR #554, opened 2016-10-24
short package desc check

gbaz:
Check for short package descs as per https://github.com/haskell/cabal/pull/3479

The version of the check via cabal won't make it into hackage for some time, but this is worth getting in place for hackage proper sooner.


gbaz:
@hvr do we want to get this deployed soon too?

Kleidukos:
@gbaz Do you need anything to land this PR?

gbaz:
I think this just got lost in the weeds.

gbaz:
Actually wait i'm pretty sure this is superceded by it being resolved  upstream in cabal (in the linked ticket). The server checks are additional checks on top of the cabal checks.

PR #576, opened 2017-03-21
Libify hackage-server.

ezyang:
I absolutely hate it when Cabal goes off and builds my modules
multiple times.  This patchset turns hackage-server into a library
so we only need to build the libraries once.

Signed-off-by: Edward Z. Yang <ezyang@mit.edu>


hackage-server.cabal
```diff
@@ -102,11 +102,8 @@ flag network-uri
   description: Get Network.URI from the network-uri package
   default: True
 
-executable hackage-server
-  if ! flag(build-hackage-server)
-    buildable: False
-  main-is: Main.hs
-  other-modules:
+library
```
hvr: please make this an internal lib!

ezyang: Only if you're all OK with only being buildable with an unreleased version of Cabal :O

hvr: I'm actually fine with it; cabal 2.0 is about to be released anyway; the travis CI job runs on cabal 2.x; and there's very few people that actually run their own hackage-server instance so there's not many ppl that will be affected anyway, and I was about to set `cabal-version:>=2.0` in the `.cabal` file anyway :-)

Issue #481, opened 2016-03-11
Uploads of new packages should go to 'candidates' by default


gbaz:
This was suggested as a change to cabal-install by @hvr but we can just change hackage-server itself to not allocate new public namespace unless a package is promoted from candidate.

This should encourage the use of the candidates feature, and more importantly vitiate against the accidental upload of internal stuff to the public hackage instance.

hvr:
Actually, I  did also suggest a change to hackage-server for namespace-allocation exactly as you describe (but maybe on a different occasion than my original cabal-install centered suggestion), since we can't easily make sure all clients are updated to the new policy (old cabal-install clients as well as 3rd party clients such as `curl`-based scripts or Stack).

gbaz:
Ok I think the change in cabal is a nicer approach, and is now done (though not in 1.24.0.0) so I'll close this in favor.

hvr:
@gbaz fwiw, it's not enough to handle this client side; we just had another accident where a package was uploaded accidentally (ghcjs-vdom) because the tooling UI was a bit error-prone and packaged up tarballs for packages that didn't belong to the user performing the upload.

hvr:
this is is somewhat related to #558

gbaz:
@hvr ok so what's the current design for this then? for gsoc we should work this out. if we just make upload url put a candidate, how do we let users force the converse? do we want to force all packages to be candidates then promoted, with no "escape hatch"?

PR #579, opened 2017-03-29
Implement upload check for #481

hvr:
This implements a very simple hack to addres #481 by
rejecting uploading the first non-candidate release
unless the user is already a member of the package
maintainers group (which candidate package users are
members as well of)

This addresses #481


3400268471:
thank

PR #606, opened 2017-08-12
Use a SVG image for the logo in the header

wereHamster:


hvr:
Thanks! May I ask where those logos are from?  I'm a bit concerned about the licence

wereHamster:
https://wiki.haskell.org/ThompsonWheelerLogo

PR #607, opened 2017-08-12
Consolidate grayscale colors

wereHamster:
The current CSS uses way too many shades of gray which are only slightly distinct. See https://build-dlcolfhjld.now.sh/#/theme/colors?a=color-plettes where I tried to collect all the colors that are used throughout the website.

### DIFF
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -587,7 +587,7 @@ ul.links li form button:hover {
   float: right;
 }
 
-strong.warning { color: red; }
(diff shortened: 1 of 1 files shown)


wereHamster:
Also replaced red with another shade of red that was already being used. All the colors which are used after these changes can be seen here: https://build-ublejijygj.now.sh/#/theme/colors. The orange and red are quite close, which is unfortunate.

There are still too many shades of gray being used IMO. But it's hard to make any further changes without knowing where/how exactly these colors are used.

PR #609, opened 2017-08-21
Extend voting interface to allow score

hvr:
This extends the package voting mechanism from boolean "stars"
to integral "lambdas", and using the statistical method outlined
in

-  http://www.evanmiller.org/how-not-to-sort-by-average-rating.html

to compute an overall score.

This has been factored out of the HSoC work #514 by @SooryaN

The code has been refactored & cleaned up by Duncan, Gershom and myself.

----

TODOs:

- [x] Validate vote-update HTTP request data (you can currently vote an arbitrary rating other than {1,2,3})

### DIFF
--- a/Distribution/Server/Features.hs
+++ b/Distribution/Server/Features.hs
@@ -257,6 +257,7 @@ initHackageFeatures env@ServerEnv{serverVerbosity = verbosity} = do
                          coreFeature
                          -- [reverse index disabled] reverseFeature
                          downloadFeature
+                         votesFeature
                          tagsFeature
                          versionsFeature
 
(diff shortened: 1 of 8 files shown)


gbaz:
lgtm, aside from the vote checking issue you mentioned. I suggest just capping scores at 1 and 3 -- i.e. normalize "too big" things rather than bothering to fail.

gbaz:
Ok, it turns out the "new auth" thing i did never worked right. Sigh. In particular, we can only ever guard and not check authentication, as browsers don't necessarily send an auth header unless challenged.

That means we can't check if a user is logged-in opportunistically on a page and display different content in either case. However, in this case we could write some further javascript to fix it up. I propose for now just deleting the new Auth stuff and the code that uses it to insert userRating into page, as its broken.

I think we can merge without it, and I can add the displaying stuff back as a feature the "right" way.

Issue #230, opened 2014-06-09
Email maintainers when the .cabal file gets edited


tibbe:
Whenever the .cabal file gets edited on Hackage we should email all maintainers (except perhaps the maintainer that made the change). This will make post-facto edits more transparent and gives us an opportunity to encourage authors to incorporate the changes upstream, so they're part of the next release.

tibbe:
I'd say this is quite high priority as otherwise there's a big risk that the changes don't get merged upstream and we have to continually fix all released versions. In addition, having the author be aware can prevent confusion when debugging issues with their packages.

hvr:
to reduce email spamming (I tend to batch-edit packages; if you add a stricter constraint, you often have to fixup previous releases as well), changes should be sent out summarised, at most once every 15 minutes or so...

bergmark:
We changed our minds and want opt-out for notifications, right?

It might also be useful to say something about the revising user e.g., "JohanTibell  (maintainer)", AdamBergmark (trustee)" so it's clear why the user has permission to create a new revision.

tfausak:
As a package maintainer, I want these notifications. If a trustee changes one of my packages on Hackage, I want to know. I also want these notifications to be opt-out rather than opt-in. Having these notifications could have prevented (or at least ameliorated) a recent dust-up regarding package revisions: https://www.reddit.com/r/haskell/comments/67u8he/stackages_norevisions_experimental_field/dgtoz9o/?context=4

hvr:
A somewhat related issue was filed by @eborden [here](https://github.com/hvr/hackage-matrix-builder/issues/33). I'd really like to see this implemented, as I'd hope this would result in more appreciation for the large amount of work Hackage Trustees are investing to help package maintainers keep the Hackage ecosystem in a healthy state, which is essential for providing a good user experience to both veterans (who need to be able to rely on Hackage & Cabal to "just work" to get work done and make productive use of Haskell) as well as newcomers (which would otherwise be discouraged before even having written the first line of Haskell) without having to rely on external 3rd-party systems.

I think for this long-standing feature-request to make progress we need to come up with an implementable specification. The email-based approach still seems to me to be the most reasonable to implement at this point (automatically filing issues or even PRs on GitHub/GitLab/DarcsHub/whatever etc would be require much more effort in terms of complexity as well as risking being in conflict with the respective terms of uses regarding spam/bot-use)

[...]

PR #613, opened 2017-08-27
Enhance Tagging feature

hvr:
- UI for proposing tag changes

- Clickable UI for accepting/rejecting tags

- Add GET method to /package/:pkgname/tags

- Categories are parsed into tags

- Tag aliasing support
  (When a trustee aliases Tag abcd -> Tag abc, all packages that were earlier
  tagged `abcd` get tagged to `abc` and any new packages tagged `abcd` get
  retagged `abc` on upload)

This has been factored out of the HSoC work #514 by @SooryaN

The code has been refactored & cleaned up by Duncan, Gershom and Herbert.

### DIFF
--- a/Distribution/Server/Features.hs
+++ b/Distribution/Server/Features.hs
@@ -241,6 +241,7 @@ initHackageFeatures env@ServerEnv{serverVerbosity = verbosity} = do
     tagsFeature     <- mkTagsFeature
                          coreFeature
                          uploadFeature
+                         usersFeature
 
     versionsFeature <- mkVersionsFeature
                          coreFeature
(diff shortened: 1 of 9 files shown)


gbaz:
Things we may want to consider doing: adding `lic:` before license tags, dropping the `test` tag as useless.

I was a bit worried about cycles in tag aliases, but after reading the code carefully and seeing how they're linearized for update, it won't actually cause problems, and the latest will just win.

Future work could include a better interface for reviewing current aliases or displaying them.

Also, future work related to https://github.com/haskell/hackage-server/issues/230 is to add notifications to maintainers for proposed tag additions, and also putting those together on a "maintainer/trustee" dashboard page whenever that's created. Absent that, proposed tags may well get lost in the ether...

PR #614, opened 2017-08-30
Read haddocks new meta.json and prepare for QuickNav

alexbiehl:
@hvr This PR teaches hackage to read haddocks new `meta.json`. `meta.json` contains useful additional information about the generated documentation (see https://github.com/haskell/haddock/pull/676). 

We use the additional information to integrate haddocks new QuickNav feature on the package page. As this integration is likely to change in the future it is guarded by the haddock version.


hvr:
This works well enough for me; I'll squash & merge this now to preserve your autorship, and give it some minor finishing touches as follow-up commit.

PR #622, opened 2017-09-19
User Notification emails

gbaz:
This is a mergable chunk that should almost resolve https://github.com/haskell/hackage-server/issues/230

It has machinery for user preferences for which notifications to get, but no UI or API for them yet. Also, by default it has user preferences set to _not_ get emails.

So it should be safe to review and merge as-is.

However, until we have a user-pref UI, nobody will be able to opt-in to it. And once that's done and we're comfortable, we can then cut over to an opt-in by default in the future.

There's also flags but no implementation work done yet for
 -- notifications on pending proposed tags
 -- notifications on docbuilder reports

As is it notifies on users being added to or removed from maintainer groups, and on metadata revisions and new package uploads -- which are the key things from a collaboration and transparency standpoint. The others, which can be in a future PR, are nice-to-have.


tfausak:
Can I do anything to help move this forward? I really want this feature.

gbaz:
I think this chunk should be reviewed and merged when possible. But it will need to be paired, as I mentioned, with a user-pref UI that needs to be written. Basically we need a page where logged-in users can modify these settings, and one hasn't been written. This is a standalone chunk of work that anyone can pick up, though it requires understanding how all the acid-state component stuff works, to a degree.

gbaz:
If we had docbuilder notifications those also could be a good place to let users know what to do in the event of a build failed due to missing system packages.

gbaz:
This is complete afaik.

Issue #615, opened 2017-09-03
Shoud fuzzy search be supported?


suzumiyasmith:
Now, it is obvious that fuzzy search is not supported. e.g. search `rethink` won't get `rethinkdb` package as a result. Sometimes, it can be annoying. :(

Any idea or plan about fuzzy searching?

Thanks.

tfausak:
I would like to see this too. I can't remember the number of times I've searched for a partial string and been surprised by the lack of results. Some recent examples that I remember:

- Searching for `influx` (https://hackage.haskell.org/packages/search?terms=influx) returns no results even though there is an `influxdb` package (https://hackage.haskell.org/package/influxdb).
- Searching for `postgres` (https://hackage.haskell.org/packages/search?terms=postgres) doesn't return `postgresql-simple` (https://hackage.haskell.org/package/postgresql-simple), which is the most popular PostgreSQL package.

That being said, I'm not holding my breath for a fix. Hackage's search appears to be based on keywords and word stems. And search issues tend to stay open for ... a while.

- https://github.com/haskell/hackage-server/issues/474
- https://github.com/haskell/hackage-server/issues/208
- https://github.com/haskell/hackage-server/issues/27
- https://github.com/haskell/hackage-server/issues/21
- https://github.com/haskell/hackage-server/issues/20

Have you tried searching on Stackage? It does a better job for all the queries given here:

- https://www.stackage.org/lts-9.3/hoogle?q=rethink
- https://www.stackage.org/lts-9.3/hoogle?q=influx
- https://www.stackage.org/lts-9.3/hoogle?q=postgres

gbaz:
Hackage is under active development at the moment, and search is definitely a place we want to tackle! thanks for the links to related issues, but I wouldn't be so negative about the possibility for progress. More hands and more hackers are always welcome :-)

gbaz:
So I think the simplest thing to do is ad-hoc improve our stemmer since it seems to work just fine except for "weird" computery stems like dropping "db" or "ql" etc.

adamgundry:
It might be worth switching Hackage to use the `full-text-search` package. This was derived from the Hackage search implementation (so it provides a very similar interface and switching should be fairly trivial), but it has a few extra features, notably autocomplete search.

gbaz:
Good idea. Any volunteers? :-P

adamgundry:
[...]

PR #633, opened 2017-09-30
Ad hoc stemming

gbaz:
should resolve the concrete issues posed in https://github.com/haskell/hackage-server/issues/615


gbaz:
Hmm, it looks like despite deploying this we're not seeing improved search results in the cases it was supposed to cover... do we need to force a rebuild of the search index, or did this just not work?

gbaz:
@hvr: if you look here you'll see that the index isn't persisted to disk but seems to be generated on startup each time. So... not sure where this patch went wrong... https://github.com/haskell/hackage-server/blob/59f95714b2e82fede85dc1de28f5043572ae95d8/Distribution/Server/Features/Search.hs

PR #647, opened 2017-12-26
Change main suggested browse index to table-based one

gbaz:
should resolve https://github.com/haskell/hackage-server/issues/326


gbaz:
lolwhoops i think this duplicates #552 which was never merged? we should merge that one.

hvr:
@gbaz ack! I'll look into this one

hvr:
done

PR #648, opened 2017-12-26
Improve appearance and readability

NunoAlexandre:
## Motive

I work with haddock and hackage almost daily, both at work and at home, and the current theme is unfriendly. I find it very hard to read content with such a wide `#content` and spacing is not great either. 

Furthermore, I'd say many Haskell developers would appreciate the presense of the latest Haskell's logo (purple) colors in these docs. 

(Together with https://github.com/haskell/haddock/pull/721)

## Changes

- use latest Haskell's logo colours
- decrease #content width to improve readability
- use, custom nicer font: [Merriweather Sans](https://fonts.google.com/specimen/Merriweather+Sans)
- improve sizes and distances
- responsiveness (the menu is now always fully visible and good looking) 

## Preview

**EDIT 04-02-2018**: See https://github.com/haskell/hackage-server/pull/648#issuecomment-362934971
<br>

**Before:**

before-hackage

**After:**

hackage-after-big

### Smartphone

**Before - iPhone 7**

hackage-before-iphone7

**After - iPhone 7**

hackage-after-iphone7

### Tablet

**Before - iPad horizontal**

hackage-before-ipad-mini

**After - iPad horizontal**

hackage-after-ipad-mini-lay

**Before - iPad vertical**

hackage-before-ipad-mini-stand

**After - iPad vertical**

hackage-after-ipad-mini-stand

## Edit: Responsiveness

I have made it responsive so that it looks nice and reads well on both small and large screens. See commit summary for more information.

### DIFF
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -51,7 +51,7 @@ a[href]:hover { text-decoration:underline; }
  */
 
 body {
-	font: 300 14px/1.85 "Merriweather Sans", sans-serif;
+	font: 400 15px/1.85 'Open Sans', sans-serif;
 	*font-size:small; /* for IE */
 	*font:x-small; /* for IE in quirks mode */
 }
@@ -218,7 +218,7 @@ pre {
     text-align: left;
     float: right;
     display: inline-table;
-    margin: 3px 2em 0 2em;
+    margin: 2px 2em 0 2em;
   }
 
   #content {
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -190,14 +190,14 @@ ul.links li a, ul.links li form {
 }
 
 pre {
-  padding: 17px;
-  margin: 1em 0 2em 0;
+  padding: 1rem;
+  margin: 0px;
   background-color: rgba(0, 0, 0, .033);
   overflow: auto;
-  border-bottom: 0.25em solid white;
-  /* white border adds some space below the box to compensate
-     for visual extra space that paragraphs have between baseline
-     and the bounding box */
+}
+
+pre + pre {
+  margin-top: 0.4em;
 }
 
 .src {
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -224,7 +224,7 @@ pre {
     text-align: left;
     float: right;
     display: inline-table;
-    margin: 2px 2em 0 2em;
+    margin: 0.35em 2em 0 2em;
   }
 
   #content {
@@ -271,10 +271,10 @@ pre {
 
 
 #page-header {
-  background: rgb(94, 81, 132);
-  border-bottom: 5px solid rgba(69, 59, 97, 0.5);
+  background: #5E5184;
+  border-bottom: 0.35em solid rgba(69, 59, 97, 0.5);
   color: #ddd;
-  padding: 0.6em 0 0.2em 0;
+  padding: 0.6em 0 0.3em 0;
   position: relative;
   font-size: 1.2em;
   margin: 0 auto;
@@ -287,7 +287,7 @@ pre {
   font-weight: bold;
   font-style: normal;
   padding-left: 2.2em;
-  font-size: 1.05rem;
+  font-size: 1rem;
 }
 
 #page-header a:link, #page-header a:visited { color: white; }
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -17,8 +17,16 @@ body {
   position: relative;
 }
 
+table, ul {
+	line-height: 1.35rem;
+}
+
+ul li + li {
+	margin-top: 0.4rem;
+}
+
 p {
-  margin: 0.8em 0;
+  margin-top: 1.1rem;
 }
 
 ul, ol {
@@ -51,7 +59,7 @@ a[href]:hover { text-decoration:underline; }
  */
 
 body {
-	font: 400 16px/1.75 'Open Sans', sans-serif;
+	font: 400 16px/1.6 'Open Sans', sans-serif;
 	*font-size:small; /* for IE */
 	*font:x-small; /* for IE in quirks mode */
 }
@@ -119,13 +127,12 @@ pre, code, kbd, samp, .src {
 
 .caption, h1, h2, h3, h4, h5, h6 {
   font-weight: bold;
-  color: rgb(142, 80, 132);
-  margin: 2em 0 1em 0;
+  color: #5E5184;
+  margin: 1.8em 0 0.95em 0;
   line-height: 1.25em;
 }
 
 h1:first-of-type {
-    color: rgb(94, 81, 132);
     font-size: 2em;
 }
 
@@ -139,7 +146,7 @@ h1 + h2, h2 + h3, h3 + h4, h4 + h5, h5 + h6 {
 }
 
 p + ul {
-  margin-top: 1em;
+  margin-top: 0.75em;
 }
 
 ul + p {
@@ -164,21 +171,21 @@ ul.links li a, ul.links li form {
 .show { display: inherit; }
 .clear { clear: both; }
 
-.collapser {
-  background-image: url(minus.gif);
-  background-repeat: no-repeat;
+.collapser:before, .expander:before {
+  font-size: 0.9em;
+  color: #5E5184;
+  display: inline-block;
+  padding-right: 7px;
 }
-.expander {
-  background-image: url(plus.gif);
-  background-repeat: no-repeat;
+
+.collapser:before {
+  content: '-'
 }
-p.caption.collapser,
-p.caption.expander {
-  background-position: 0 0.4em;
+.expander:before {
+  content: "+";
 }
+
 .collapser, .expander {
-  padding-left: 14px;
-  margin-left: -14px;
   cursor: pointer;
 }
 
@@ -277,9 +284,10 @@ pre {
   background: url(icons/ic_haskell_grayscale_32.svg) no-repeat 0em;
   color: white;
   margin: 0 2em;
-  font-weight: normal;
+  font-weight: bold;
   font-style: normal;
   padding-left: 2.2em;
+  font-size: 1.05rem;
 }
 
 #page-header a:link, #page-header a:visited { color: white; }
@@ -433,14 +441,6 @@ ul.links li form button {
   margin: 0;
   font-size: 1px;
   padding: 0;
-}
-
-#synopsis p.caption.collapser {
-  background: url(synopsis.png) no-repeat -64px -8px;
-}
-
-#synopsis p.caption.expander {
-  background: url(synopsis.png) no-repeat 0px -8px;
 }
 
 #synopsis ul {
@@ -715,10 +715,6 @@ table.simpletable th, table.simpletable td {
   padding: 0.2em 1em;
 }
 
-table.properties td, table.properties th {
-  padding: 1px 0.25em;
-}
-
 table.properties td:first-child, table.properties th:first-child {
   padding-left: 0;
 }
@@ -746,12 +742,14 @@ table.fancy th {
   background: #f0f0f0;
 }
 
-table.fancy td, table.fancy th {
+table.fancy td, table.properties td,
+table.fancy th, table.properties th {
   padding: 0.25em 0.5em;
 }
 
-table {
-  line-height: 1.5em;
+table.properties tr th:first-child {
+  text-align: right;
+  padding-right: 12px;
 }
 
 table.dataTable.compact.fancy tbody th,
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -749,6 +749,15 @@ table.fancy td, table.fancy th {
   padding: 0.25em 0.5em;
 }
 
+table {
+  line-height: 1.5em;
+}
+
+table.dataTable.compact.fancy tbody th,
+table.dataTable.compact.fancy tbody td {
+  padding: 6px 10px;
+}
+
 ul.searchresults li {
   margin-bottom: 1em;
 }
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -222,7 +222,8 @@ pre {
 
   #content {
     width: 55vw;
-    max-width: 1200px;
+    max-width: 1450px;
+    min-width: 830px;
   }
 }
 
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -15,6 +15,12 @@ body {
   text-align: left;
   min-height: 100%;
   position: relative;
+  -webkit-text-size-adjust: 100%;
+  -webkit-font-feature-settings: "kern" 1;
+  -moz-font-feature-settings: "kern" 1;
+  -o-font-feature-settings: "kern" 1;
+  font-feature-settings: "kern" 1;
+  font-kerning: normal;
 }
 
 table, ul {
@@ -89,6 +95,10 @@ pre, code, kbd, samp, .src {
 	line-height: 1.54em;
 }
 
+#page-header .links {
+  font-size: 1rem;
+}
+
 .links, .link {
   font-size: 85%; /* 11pt */
 }
@@ -114,10 +124,6 @@ pre, code, kbd, samp, .src {
     border-left: 6px solid #2196F3;
     padding: 0.01em 16px;
     margin-bottom: 8px;
-}
-
-#table-of-contents, #synopsis  {
-  /* font-size: 85%; /* 11pt */
 }
 
 
@@ -225,6 +231,13 @@ pre + pre {
     float: right;
     display: inline-table;
     margin: 0.35em 2em 0 2em;
+  }
+
+  #table-of-contents {
+    position: fixed;
+    left: 10px;
+    max-width: 10vw;
+    top: 10.2em;
   }
 
   #content {
@@ -391,18 +404,15 @@ ul.links li form button {
 /* @group Front Matter */
 
 #table-of-contents {
-  float: right;
-  clear: right;
-  background: #faf9dc;
-  border: 1px solid #d8d7ad;
-  padding: 0.5em 1em;
-  max-width: 20em;
-  margin: 0.5em 0 1em 1em;
+  background:  #f7f7f7;
+  padding: 1em;
+  margin: 1em 0 2em 0;
 }
 
 #table-of-contents .caption {
   text-align: left;
   margin: 0;
+  font-size: 1.08rem;
 }
 
 #table-of-contents ul {
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -121,6 +121,7 @@ pre, code, kbd, samp, .src {
   font-weight: bold;
   color: rgb(142, 80, 132);
   margin: 2em 0 1em 0;
+  line-height: 1.25em;
 }
 
 h1:first-of-type {
@@ -143,14 +144,6 @@ p + ul {
 
 ul + p {
   margin-top: 2em;
-}
-
-ul.links {
-  list-style: none;
-  text-align: left;
-  float: right;
-  display: inline-table;
-  margin: 0 2em 0 2em;
 }
 
 ul.links li {
@@ -211,14 +204,57 @@ pre {
 
 /* @end */
 
+
+/* @group responsive */
+
+@media only screen and (min-width: 950px) {
+  #page-header {
+    text-align: left;
+    text-align: left;
+  }
+
+  #page-header ul.links {
+    list-style: none;
+    text-align: left;
+    float: right;
+    display: inline-table;
+    margin: 3px 2em 0 2em;
+  }
+
+  #content {
+    width: 55vw;
+  }
+}
+
+@media only screen and (max-width: 950px) {
+  #page-header {
+    text-align: center;
+  }
+
+  #page-header ul.links {
+    float: none;
+    display: block;
+    text-align: center;
+    margin: 14px 0 0px 0;
+    font-size: 13px;
+  }
+
+  #content {
+    width: 75vw;
+  }
+}
+
+
+/* @end */
+
 /* @group Page Structure */
 
 #content {
   margin: 0 auto;
   padding: 0;
-  width: 55vw;
   padding-bottom: 3em;
 }
+
 
 #page-header {
   background: rgb(94, 81, 132);
@@ -226,7 +262,6 @@ pre {
   color: #ddd;
   padding: 0.6em 0 0.2em 0;
   position: relative;
-  text-align: left;
   font-size: 1.2em;
   margin: 0 auto;
 }
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -64,10 +64,10 @@ a[href]:hover { text-decoration:underline; }
       http://yui.yahooapis.com/3.1.1/build/cssfonts/fonts.css
  */
 
-body {
-	font: 400 16px/1.6 'Open Sans', sans-serif;
-	*font-size:small; /* for IE */
-	*font:x-small; /* for IE in quirks mode */
+body, button {
+  font: 400 15px/1.5 'Open Sans', sans-serif;
+  *font-size:small; /* for IE */
+  *font:x-small; /* for IE in quirks mode */
 }
 
 h1 { font-size: 153.9%; /* 20px */ }
@@ -78,9 +78,9 @@ h5 { font-size: 100%;   /* 13px */ }
 h6 { font-size: 100%;   /* 13px */ }
 
 select, input, button, textarea {
-	font:99% sans-serif;
-    margin: 0.5em;
-    padding: 0.1em;
+  font-size: 1rem;
+  margin: 0.5em;
+  padding: 0.1em;
 }
 
 table {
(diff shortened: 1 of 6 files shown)


gbaz:
nice!

NunoAlexandre:
Thank you, @gbaz.

I've now made it responsive. Can you have another look at the preview screenshots above?
Thanks!

hvr:
I'm a bit confused tbh :-) Why is it considered a good thing that the new stylesheet seems to waste space left/right on a mobile device? Wouldn't you rather want to exploit the little screen estate you have on a mobile device as much as possible?

NunoAlexandre:
Hi @hvr,

> Why is it considered a good thing that the new stylesheet seems to waste space left/right on a mobile device

"Waste" is subjective, but you have a fair point here. Design and typography wise, the empty space, also called ["negative space"](http://blog.teamtreehouse.com/white-space-in-web-design-what-it-is-and-why-you-should-use-it) is beneficial, and on mobile devices, I have already reduced the "wasted" space. Maybe you mean that you still find it too much? I am fine with reducing it.

>  Similarly, on desktops; I have a rather crowded tiled WM, and I make my browser windows rather small; so I'd expect it to use the little window space as efficiently as possible to display the content.

It's not so much about if it's on mobile or desktop, but about how much space is available. This is solved by the previous point.

I also use a WM (i3wm), and I'd rather not have content taking the 100% of the available space as that is a disaster in terms of content proximity, and it's hard to read let alone.

hvr:
@NunoAlexandre Fwiw, I've given it a quick test-drive; it's not as bad as I thought when I make my windows smaller; there's a size at which the negative space suddenly snaps to a smaller amount, so i guess that's good enough :-)

However, I have a high DPI display (somewhere around 120 DPI), and while I can read e.g. GitHub's fonts in its UI just fine (without needing to zoom in), I'm feeling eyestrain with the font style and lower contrast of your stylesheet. Is there something that can be done to improve readability? 


Btw, just in case you haven't seen this yet, there's also coloured blocks like e.g. at http://hackage.haskell.org/package/bad-package-example-0/candidate ... which design-wise may pose a challenge :-)

NunoAlexandre:
> @NunoAlexandre Fwiw, I've given it a quick test-drive; it's not as bad as I thought when I make my windows smaller; there's a size at which the negative space suddenly snaps to a smaller amount, so i guess that's good enough :-)

Thanks for giving it a spin and for the feedback 👍 

> However, I have a high DPI display (somewhere around 120 DPI), and while I can read e.g. GitHub's fonts in its UI just fine (without needing to zoom in), I'm feeling eyestrain with the font style and lower contrast of your stylesheet. Is there something that can be done to improve readability?

This is good information. Mine is a 227 ppi display and I didn't notice the things you mentioned. Could you perhaps attach a screenshot? We could, for example, darken the colour of the text a little and/or reconsider the font. I like this one quite a lot because it's both very elegant and discrete. I.e, it's not just Helvetica again, which would bring a special feeling to hackage, without being an intrusive font that drags the reader away from the content. But we could try `Open Sans` or `Source Sans Pro`, for example:


### Current (Merriweather Sans)

hackage-after-big


### Open Sans

open sans


### Source Sans Pro
source sans pro

NunoAlexandre:
> Btw, just in case you haven't seen this yet, there's also coloured blocks like e.g. at http://hackage.haskell.org/package/bad-package-example-0/candidate ... which design-wise may pose a challenge :-)

I'm not sure what you meant by "challenge" in this case. It looks alright to me:

screen shot 2017-12-28 at 12 38 25

What I don't like is to have the title in the header. It's a bad idea, imo, as its size is always different, which makes the header size unpredictable and ugly often enough. I'd rather move the title to the #content, and let it always be `Hackage :: [Package]` in the header.

What do you think?

phadej:
Can we have a screenshot of https://hackage.haskell.org/package/lens-4.16/docs/Control-Lens-Operators.html with new stylesheet. It feels that there is more vertical whitespace around/in code blocks, but I'm not sure how bad it is on pages having a lot of doctests and other code blocks?

TikhonJelvis:
Looks great. I think the extra whitespace is great—having shorter lines of text is much easier to read.

In the screenshots the purple text looks a bit low-contrast, but that might be because of how the images are resized. The full-size images look better. Might still be worth making the text a bit darker.

NunoAlexandre:
> Can we have a screenshot of https://hackage.haskell.org/package/lens-4.16/docs/Control-Lens-Operators.html with new stylesheet. It feels that there is more vertical whitespace around/in code blocks, but I'm not sure how bad it is on pages having a lot of doctests and other code blocks? @phadej 

There is an [open PR on the haddock repository](https://github.com/haskell/haddock/pull/721) with the same changes proposed here. The page of your link will be affected by that PR, not this one. I will add the screenshot you ask for over there.

NunoAlexandre:
> Might still be worth making the text a bit darker. @TikhonJelvis 

Agreed. @hvr raised the same issue. I was using Merriweather Sans, which is a low-contrast font. I switched now to Open Sans, which improves that issue. I will post some more screenshots tomorrow.

NunoAlexandre:
I have made the text color darker and changed the link colors from shades of orange to shades of purple.

Here follow two samples:
hackage-home
monoid-extras

gbaz:
We need to handle preferred/non-preferred/deprecated in whatever scheme is done, and update the documentation accordingly.

cf https://github.com/haskell/hackage-server/pull/373

and for the docs: https://github.com/haskell/hackage-server/blob/2e70ec313379d75ede6c70ec211b4551777ec80b/Distribution/Server/Features/Html.hs#L1368

Other than that, this is looking pretty good!

cocreature:
> We need to handle preferred/non-preferred/deprecated in whatever scheme is done, and update the documentation accordingly.

Let’s take this opportunity to switch to a color scheme that is intuitively understandable. Green should not signify “errors”.

moodmosaic:
The current stylesheet is more readable on mobile devices. I'm actually with @hvr here:

>Why is it considered a good thing that the new stylesheet seems to waste space left/right on a mobile device? Wouldn't you rather want to exploit the little screen estate you have on a mobile device as much as possible?

---

I'd rather just apply the colors of the proposed stylesheet to the old one; and I'll leave the rest as-is.

gbaz:
The mobile margins are tighter than the desktop margins, and seem sort of in-line with how other mobile-optimized sites tend to do it.

I like high information-density but I'm sympathetic towards moving more towards how "everyone else" does it with regards to mobile these days, which is emphasizing readability over density to a certain degree.

As a whole this is _definitely_ an improvement, and I'd like to see the preferred/etc colors sorted out so we can move forward with a merge while we have good momentum here :-)

NunoAlexandre:
@moodmosaic 

You might not have seen this part:

> @NunoAlexandre Fwiw, I've given it a quick test-drive; it's not as bad as I thought when I make my windows smaller; there's a size at which the negative space suddenly snaps to a smaller amount, so i guess that's good enough :-)

In any case, that space can obviously be tweaked.

datafiles/static/hackage.css
```diff
@@ -192,24 +202,68 @@ pre {
...
 
+/* @end */
+
+
+/* @group responsive */
+
+@media only screen and (min-width: 950px) {
+  #page-header {
+    text-align: left;
+    text-align: left;
```
phadej: double?

datafiles/static/hackage.css
```diff
@@ -192,24 +202,68 @@ pre {
...
+  #page-header ul.links {
+    list-style: none;
+    text-align: left;
+    float: right;
+    display: inline-table;
+    margin: 2px 2em 0 2em;
+  }
+
+  #content {
+    width: 55vw;
```
phadej: as this is narrow page, should that be closer to `100vw`?

datafiles/static/hackage.css
```diff
@@ -192,24 +202,68 @@ pre {
...
+  #page-header ul.links {
+    float: none;
+    display: block;
+    text-align: center;
+    margin: 14px 0 0px 0;
+    font-size: 15px;
+  }
+
+  #content {
+    width: 75vw;
```
phadej: should we however have also `max-width: 1200px` or something like that? Some (accidentally) full screen pages on ridiculously large displays :)

phadej: Also, imho we *should* have `min-width` which is at wide as narrow viewport close to the breakpoint.

FWIW, Iphone7+ is 1920-by-1080-pixel resolution at 401 ppi, i.e. 960px virtual pixels in landscape, which is wide viewport (as the breakpoint is at 950px).

vagarenko:
@NunoAlexandre 
>changed the link colors from shades of orange to shades of purple

Personally, I don't like new colors. I think the current Hackage color scheme is perfect. It's high contrast and has ideal warm-cool color balance, unlike dull grey monochrome Stackage-like scheme you moving towards.

Compare 

https://hackage.haskell.org/package/lens-4.16/docs/Control-Lens-Operators.html

https://www.stackage.org/haddock/lts-10.4/lens-4.15.4/Control-Lens-Operators.html

IMO, the Hackage version is a lot more readable and pleasant to work with.

datafiles/static/hackage.css
NunoAlexandre: Yes, close to, to keep a minimally acceptable margin.

datafiles/static/hackage.css
NunoAlexandre: I have added the max-width you suggested. Thanks!

I am not sure about the min-width though. Having a page that floats around as one scroll is annoying, and it doesn't really help reading it. Do you guys have analytics data about how many people use hackage using a small device?

elben:
@NunoAlexandre I appreciate the work you've set out to do here. Knowing the risks of "design feedback via committee", here are some of my observations:

- I think we need to consider for two different needs and reading experiences. Some pages are intended for a thorough read (e.g. tutorials) where readability and flow is important. Others like Haddock pages are references where density is more important than readability.

- So I don't think making things less-dense is a necessarily a good thing. I don't understand this design trend of massive negative space. It may be great for Medium articles, but not for every kind of material. I think it is the wrong approach to say "we need more negative space" (which may be valid), and apply it everywhere to this extent.

- What's up with the low-contrast trend on web design? I can barely see the text. You can't imagine any published book having a contrast ratio so light.

- One thing I appreciate today about the Haddock pages is that I can see many functions at once on a single page. I can compare functions against each other, without having to massively scroll all over the page.

- I think your line-height is a bit too much. Notice how awkward the breaks between lines and paragraphs are below. They are too-similarly spaced:

screen shot 2018-02-01 at 2 47 08 pm

NunoAlexandre:
### Update

- Adjust  responsive behaviour
    - Have special sizes for small, medium and large screens.
    - Now the space is further exploited in small screens.
    - Set a max-width

<details>
<summary>Screenshots</summary>
- Iphone7s
iphone7s

- iPad Pro
ipadpro

- Desktop
desktop

- Vertical (using i3wm)
vertical


</details>


- Make package version color scheme intuitive (see https://github.com/haskell/hackage-server/pull/648#issuecomment-361396640)
The current color scheme is untuitive. Also, add the class "versions" to
the element holding the list of versions of a given package, and add
the class "normal" to normal versions. This helps making the style more
specific and explicit, while being able to drop the "!important" flags.

<details>
<summary>Screenshot</summary>

colors

</details>

gbaz:
This is really coming along! It seems the remaining issues around readability mainly involve haddocks rather than the core hackage pages. Perhaps it could be useful to pursue discussions regarding them on that PR?  (https://github.com/haskell/haddock/pull/721). I think just bumping contrast a bit on the purple headers could help a ton.

(btw do you have screenshots of the table-based results that come from the /browse page etc? It would be nice to see.)

NunoAlexandre:
Hi @elben, thank you for your words and for your constructive feedback.


> I think we need to consider for two different needs and reading experiences. Some pages are intended for a thorough read (e.g. tutorials) where readability and flow is important. Others like Haddock pages are references where density is more important than readability.

I personally cannot recall a case where density was more important than readability, for I would not prefer to see more content at once and have a much harder time reading any of it. I might be wrong here, I am not dismissing your observation. 

> So I don't think making things less-dense is a necessarily a good thing. I don't understand this design trend of massive negative space. It may be great for Medium articles, but not for every kind of material. I think it is the wrong approach to say "we need more negative space" (which may be valid), and apply it everywhere to this extent.

Do you have an example of a material which is better read with a dense design than with a good negative space that you can show? 

I am not with you on this one. I have minor studies in typography and (negative) space is fundamental, and a lot of the things you are not into or don't understand have been thoroughly studied and tested. You are obviously free to have your preference, but I am afraid your are not on the winning side of the discussion.

If your point is that I am using too much negative space now, I am open to improve it. But I'd rather have a first version deployed and see what the masses have to say and tweak accordingly, as now the majority seems to like it as I am proposing (which obviously doesn't mean it's right), and democracy rules. 

> What's up with the low-contrast trend on web design? I can barely see the text. You can't imagine any published book having a contrast ratio so light.

You can't really compare screens with books.
I agree there is an exageration of low-contrast overall on the web, but I don't see how that is the case here? I think an example of it is, actually, Stackage (which I prefer to look at than Hackage, FWIW), but - again - I can't see how that is a thing on this PR.

> One thing I appreciate today about the Haddock pages is that I can see many functions at once on a single page. I can compare functions against each other, without having to massively scroll all over the page.

If what you want is to look at raw code, why not look at the source? It gives your all you need and extras. However, we can indeed think of some code-reading mode, but let's not go off-scope here. 

> I think your line-height is a bit too much. Notice how awkward the breaks between lines and paragraphs are below. They are too-similarly spaced:

I agree, thanks!

NunoAlexandre:
Hi @gbaz,

> This is really coming along!
Cool, nice getting your feedback!

> It seems the remaining issues around readability mainly involve haddocks rather than the core hackage pages. Perhaps it could be useful to pursue discussions regarding them on that PR? (haskell/haddock#721). 

Correct. I have delayed working on haddock for 1) lack of time to work on both and 2) to avoid having to make the same changes in both places. Once we are done here I will move to there.

> I think just bumping contrast a bit on the purple headers could help a ton.

Are you referring to the top header (menu) or to the headings (titles)?

> (btw do you have screenshots of the table-based results that come from the /browse page etc? It would be nice to see.)

Sure, here it is:
table-hackage

gbaz:
> Are you referring to the top header (menu) or to the headings (titles)?

I was referring to the headings, and more in the haddock case actually...

On the table view, I was wondering if you, as a designer, had any thoughts on adding a bit of extra data. I tried tooltips and wasn't super-happy with the result. c.f. https://github.com/haskell/hackage-server/pull/663

It would be nice to unobtrusively specify that the downloads are 30-day counts and the ratings are 0-3.

moodmosaic:
Like @vagarenkos [wrote](https://github.com/haskell/hackage-server/pull/648#issuecomment-362360431), I also think that the current theme is fine.

*Not* that the proposed theme isn't cool... it's just a matter of taste. But if you've used Hackage for years, you might want to have at least the option to switch back to the 'classic' theme.

phadej:
About `min-width` for large viewport, look at this screenshots:

## narrow viewport

screenshot from 2018-02-02 15-39-03

## wide viewport, content is narrower

This is IMO bad!

screenshot from 2018-02-02 15-39-26

--- 

I made a small adjustments, see http://oleg.fi/lens-haddocks/

There the narrow-wide breakpoint is at 960px

The above bug doesn't happen, *content doesn't ever shrink when you widen the viewport*.

It also uses full width on my iphone 7 landscape and portrait, so I'm happy. :)

I hope that my changes are accepted by @NunoAlexandre and we can merge responsiveness aspect, and discuss about typography separately?

```diff
+/* @group responsive */
+
+/* Wide screen */
+@media only screen and (min-width: 960px) {
+  #page-header {
+    text-align: left;
+    text-align: left;
+  }
+
+  #content {
+    width: 60vw;
+   min-width: 864px; /* 960px * 90%, minimum value is widest narrow can become */
+   max-width: 1600px; /* prevents content to go super-wide */
+  }
+
+  #package-header .caption {
+    margin: 0 0 0 20vw;
+  }
+
+  ul.links {
+    margin: 0px 20vw 0 0;
+  }
+
+}
+
+/* Narrow screen */
+@media only screen and (max-width: 960px) {
+  #page-header {
+    text-align: center;
+  }
+
+  #content {
+    width: 90vw;
+  }
+
+  #package-header .caption {
+    margin: 0 0 0 10vw;
+  }
+
+  ul.links {
+    margin: 0px 10vw 0 0;
+  }
+}
+
+/* Very narrow screen */
+@media only screen and (max-width: 500px) {
+  #module-header table.info {
+    float: none;
+    top: 0;
+    margin: 0 auto;
+    overflow: hidden;
+    max-width: 80vw;
+  }
+}
+
+/* @end */
+
```

NunoAlexandre:
@gbaz, @phadej, and @elben 

I have addressed your points about heading colors, min-width and text line-length. See the screenshots below.

<details>
<summary>Screenshots</summary>

**Text line-length**
lens copy

**Table styling**
browse

**Heading colors**
home

</details>
<br>

@gbaz
> On the table view, I was wondering if you, as a designer, had any thoughts on adding a bit of extra data. I tried tooltips and wasn't super-happy with the result. c.f. #663 (@gbaz)

Note that I am not a designer, professionally speaking. I am keen and did minor some studies on it, but I am a software engineer working with Haskell daily. Now answering your question, I think the tables don't look very nice, but I rarely look at the `browse` page to be honest. 
 
@phadej,

I didn't quite get what you meant to show with the screenshots. And I think you haven't noticed before that comment that I introduced one more responsive stage. I have addressed your point about the min-width by setting it to 830px for larger screens. The smaller screens already make good use of the space.

phadej:
Where from 830 comes? it's not 1280 * 0.75 = 960, (950 * 0.86 is 836).

IMO the responsiveness should be outlined (in comment for future tweakers), and implemented as:

| viewport  | max viewport | content width ratio | min content width | max content width |
| --------- |--------------|---------------------|-------------------|-------------------|
| narrowest |       w1     |          r1         |       minw        |      w1 * r1  (1) |
| narrow    |       w2     |          r2         |     w1 * r1  (1)  |      w2 * r2  (2) |
| wide      |      inf     |          r3         |     w2 * r2  (2)  |        maxw       |

So there is 7 variables:
- breakpoints `w1` and `w2` (now: 950px, 1280px)
- relative content width for each viewport (now: 88%, 75%, 55%)
- absolute minimum and maximum content widths, `minw` and `maxw` (now: effectively 0, and 1450px)

I have no strong opinions on what the actual values should be, but on their relations.

Note: how (1) and (2) values come in pairs.  For example, we define *min
content width* for *narrow* viewport to to be `w1 * r1`, i.e. the *max content
width* of *narrowest* layout can get etc.  And if we do so, resizing the
viewport won't make content width "jump" wider or narrower.

NunoAlexandre:
> Where from 830 comes? it's not 1280 * 0.75 = 960, (950 * 0.86 is 836).

It comes from having tried `min-width: 950px` on my laptop and having highly dislike it because it is too wide for my screen size. `1280*0.55` = `704`. `830` was a friendly value between `704` and `950`.

 I am understanding of your point about the min-width for larger screens, but not for the others. The `narrow` and `narrowest` screens already exploit the space well enough. I gave a look at websites with nice responsive behavior and didn't find one with min-width. 


> absolute minimum and maximum content widths, minw and maxw (now: effectively 0, and 1450px)

"effectively 0" assumes that people with a screen size of a human nail will be looking at hackage. Let's not spend time on 0.00000009%  of the use cases. For larger screens, the minimum is now `830px` and max `1450px`. It surely doesn't need to follow any rigid formula.

NunoAlexandre:
# Preview latest suggestions

Instead of posting screenshots, I have hosted the pages you guys have shown interest in seeing the latest suggestions here proposed. 

You can find a list of links to them following the link below: 
https://nunoalexandre.com/2018/02/04/redesigning-haskell-docs#preview

This allows anyone to preview and try it out! I will keep on updating it as we ago so it will always be updated.

identicalsnowflake:
You can fix the header spacing like this:

- `#page-header`
    - Remove `border-bottom`
    - Set `padding` to `0.6em 0`
- `#page-header ul.links`
    - Set `margin` to `6px 2em 0 2em`

This will put the same amount of space above and below the items in the header and align all the text to the same horizontal line.

v0d1ch:
Great work! The only thing that I think could be better is to use dropdown menu for small screens but I guess you are just changing the css, not adding any extra html elements. 
Looks a lot nicer now IMO.

NunoAlexandre:
@v0d1ch 

> Great work! The only thing that I think could be better is to use dropdown menu for small screens but I guess you are just changing the css, not adding any extra html elements.
Looks a lot nicer now IMO.

Thank you very much. That is a good suggestion, but I am trying to reduce html changes as much as possible in this PR. Maybe you can create an issue here about it? The scope for now is to have the menu looking acceptable and working in small devices, which is the case now.

thumphries:
Just wanted to chime in on the contrast. The purple text against grey or white background was quite hard for me to read, so I tested it for WCAG compliance and found it falls short of AAA readability by a couple of points: https://webaim.org/resources/contrastchecker/?fcolor=9C5791&bcolor=F6F6F6 https://webaim.org/resources/contrastchecker/?fcolor=9C5791&bcolor=FEFEFE

Darkening a few shades to achieve WCAG AAA readability might be worthwhile; it would certainly stop me from immediately installing custom CSS.

v0d1ch:
@NunoAlexandre 
> Maybe you can create an issue here about it?

will do

scott-fleischman:
I like the direction this style is going.

One minor point: there's too much initial vertical space when expanding individual instances.

NunoAlexandre:
### Update
I have addressed the many comments about information density by reducing the font-size a little and improving some spacing between elements.

NunoAlexandre:
> Just wanted to chime in on the contrast. The purple text against grey or white background was quite hard for me to read, so I tested it for WCAG compliance and found it falls short of AAA readability by a couple of points:

I don't empathize with this but if you suggest alternative colors I am open to adjusting. I am not of the opinion that every single element or piece of information needs to have the highest level of contrast possible. Sometimes you just need them to be easily spotted, for example. I ran a few tests using color and contrast schemes used in some docs by Google and the whole thing fails altogether, while those are one of the most readable docs around.

thumphries:
> if you suggest alternative colors I am open to adjusting

Moving those sliders until the contrast ratio hits 7 (AAA readability) produces #6F3E69 and #794470, both of which no longer hurt my eyes to read.

> I am not of the opinion that every single element or piece of information needs to have the highest level of contrast possible.

Your opinion here shouldn't matter. Accessibility should not be optional for a widely used technical reference. Other sites may not comply, but we could do so at no cost.

TikhonJelvis:
> Your opinion here shouldn't matter.

This is a brusque way of putting it, but the core idea is sound. WCAG gives us a set of simple, objective rules for making our system useful and comfortable for as many people as possible. With contrast in particular, the standards exist to keep text readable even for people who do not distinguish colors well.

Since we aren't experts in accessibility, the guidelines help us do a good job without needing to second-guess other organizations or to generalize from our own limited experiences.

> ...that every single element or piece of information needs to have the highest level of contrast possible.

To be clear, the contrast guidelines do *not* expect *everything* to have the same level of contrast. There are different levels for different text sizes, and even the AAA standards for text below 18 pt allows a wide range of contrasts. It just provides a *minimum* contrast level chosen to work for people with 20/80 visual acuity:

> A user with 20/40 would thus require a contrast ratio of 3 * 1.5 = 4.5 to 1. Following analogous empirical findings and the same logic, the user with 20/80 visual acuity would require contrast of about 7:1.

(From [WCAG](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast7.html).)

NunoAlexandre:
I have increased the contrast of the links. IMO, now it all looks almost black. I.e, the contrast between different things is low., while it was clear before How do you go about that, @thumphries and @TikhonJelvis? I do not want to put aesthetics before accessibility, but this seems a mix of both.

fredefox:
I agree that the contrast between the text and the background should be large enough to make the text legible - but the same should go for the contrast between normal text and link text. Links are virtually indistinguishable at the moment. Perhaps saturating the color and increasing the brightness could satisfy both requirements.

gbaz:
I've got a branch here: https://github.com/haskell/hackage-server/tree/gb/stylesheet-mar18

It has a few more small changes on top, to resolve the remaining issues. It tightens some spacing around headers a bit more (and distributes it around paragraphs again, instead of putting it all at the top). Additionally, it fixes the header styles for the old (non-template) pages. It makes the links more contrasty again, and makes unvisited brighter and more saturated than visited, as is general practice. It lets the width of the browse tables grow as the page is increased. And finally, it fixes a bug where long package-descriptions could force the menu offscreen.

I feel this is mergable.

The drawback is that, if we merge now, then the haddocks will still take some time to catch-up, and in the meantime the styles will be nonuniform between haddocks and main hackage.

On the other hand, this has been underway for a few months, and it would be good to merge before momentum is lost. We can always continue to tweak styles later, so I see little downside.

Thoughts?

gbaz:
Added the two-col changes on top of this since they were very popular:

image

gbaz:
Added the two-col changes on top of this since they were very popular

image

TikhonJelvis:
Merging this now is definitely a good idea, even if it leads to some temporary inconsistency in style. I don't see anything fundamentally blocking this change and changes end up dying after sitting too long as PRs all too often...

gbaz:
Ok, good to see the sentiment. Closing this in favor of #693, which has all this plus further changes, as described above.

xerus2000:
idk if I missed something in this discussion, but why were the changes to the navigation header discarded? The style implemented by @NunoAlexandre seemed much more clean and legible. I don't understand why there's an extra blue bar at the top and stuff.

xerus2000:
oh sorry I'm stupid, I was in the haddock-part :sweat_smile:

PR #663, opened 2018-01-31
table tooltips

### LINKED ISSUE #642: downloads count in search results only show 30 day downloads
we should change the text or the contents so it matches

gbaz:
sort of resolves #642 

Adding "(30 days)" after "DLs" makes the table header unwieldy, so I addyed it as a tooltip. But that's really non-obvious to find. While I was at it I also added a tooltip to indicate ratings are 0-3.

But it would be better to have either A) better tooltips (via fancy css, I guess) or B) bite the bullet and go to 2-row headers or C) something else, I'm not sure what.

This certainly improves things slightly, but I'm not sure if it is enough :-/

### DIFF
--- a/datafiles/templates/Html/table-interface.html.st
+++ b/datafiles/templates/Html/table-interface.html.st
@@ -19,8 +19,8 @@
                 <thead>
                     <tr>
                         <th><div style="width:100px">Name</div></th>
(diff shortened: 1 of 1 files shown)


hvr:
another option would be adding some side-legend or a paragraph somewhere....

PR #670, opened 2018-02-11
Use non-breaking-space in package metadata table

phadej:
Source repository is currently split on two lines making table look "wrong". The name is long, but maybe this is till better:

screenshot from 2018-02-11 17-05-06

### DIFF
--- a/datafiles/templates/Html/package-page.html.st
+++ b/datafiles/templates/Html/package-page.html.st
@@ -58,7 +58,7 @@
 
         $if(package.optional.hasChangelog)$
         <tr>
-          <th>Change log</th>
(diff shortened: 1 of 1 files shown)


phadej:
Currently:

screenshot from 2018-02-11 17-10-30

gbaz:
What about just changing the text to “Source repo”?

phadej:
Note that it's not always singular, so if we change the name, we could make it more correct to:
screenshot from 2018-02-11 18-22-31

(with multiple repositories the word wrap isn't annoying anymore, btw)

hvr:
Let's try this one out for a while :-)

Issue #677, opened 2018-02-19
Implement Uncurated Hackage Layer


gbaz:
This issue tracks progress made on https://github.com/gbaz/ecosystem-proposals/blob/gbaz-uncurated/proposals/0000-uncurated-layer.rst

- [ ] Documentation change to upload page
- [ ] Hackage will ensure that no revision has x-curation set to any variant of "uncurated".
- [ ] implementation of UI (display curation flag in package pages and browse/search results)
- [ ] (Related: add filtering controls to browse/search results)
- [ ] Provide uncurated package repo root
- [ ] Give curated -> uncurated transitive warnings
- [ ] Enforcement of transitive deps check on upload of curated packages

gbaz:
Step one has a PR at https://github.com/haskell/hackage-server/pull/676

gbaz:
notes on ui display: for the package page we should have something like

curation: uncurated, curation: uncurated-no-contact. curation: curated. curation: curated (self-adopted). curation: adopted (by: trustee-name)

PR #679, opened 2018-02-19
check revisions to ensure no curation flags exist

gbaz:
part of https://github.com/haskell/hackage-server/issues/677


bergmark:
Let me check if I understand the intended workflow here.

* A revision can be made  to any package
* The first revision will remove the x-curation field
* For any revision made, trustees should check the original cabal file so they know if they should contact the maintainer?

Sounds good, assuming I got it right :-)

gbaz:
Close -- but one sight amendment. This PR enforces that the first revision must remove the field -- it does not automatically remove the field. 

For now, yes, trustees will need to check the original file. However, it would be appropriate for that information (curation status and history) to be placed directly in the Hackage UI, and that's on the roadmap. At that point, the information will be accessible more directly.

PR #681, opened 2018-02-21
Add GA Tag

### LINKED ISSUE #665: Add google-analytics
It would be very good to have some objective data on browsing patterns, to see how people tend to use the site. This would really inform future discussions.

I just  registered us a key  under haskell.org -- we just need to add it to all the pages.

Code follows (to be inserted in HEAD section):

```
<!-- Global site tag (gtag.js) - Google Analytics -->
<script async src="https://www.googletagmanager.com/gtag/js?id=UA-83290513-3"></script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());

  gtag('config', 'UA-83290513-3');
</script>

```

gbaz:
resolves #665


gbaz:
Whoops, this should only be against the deploy branch.

PR #687, opened 2018-03-07
Increase visibility of note about sending an email to active accounts

cocreature:
I have seen multiple people complain in IRC that they are unable to upload packages over the last few days since they missed the note on the website that they need to send an email to active their accounts. Looking at the registration page, I would probably have missed it myself.

This PR tries to address this by moving the note above the registration button (nobody reads past that) and adding a red border around it. I’m sure this could be improved but in this case I think it’s more useful to address this quickly and then bikeshed the design later (iiuc this is only intended as a temporary measure anyway) than discussing the shade of red used in the border. :wink: 

Here’s a screenshot:

screenshot-2018-3-7 register a new account hackage

### DIFF
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -722,3 +722,8 @@ ul.directory-list {
 .unpreferred {
     color:#308014 !important;
 }
+
+p.registration-email {
+    border: 1px solid rgb(196,69,29);
+    padding: 0 1em;
+}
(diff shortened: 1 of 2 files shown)


cocreature:
Pinging @gbaz, you mentioned on IRC that this looks good to you so can we get it merged and deployed somewhat soon to hopefully reduce the number of people confused by this?

hvr:
If Gershom agrees, we can definitely have this live real soon; I'm currently preparing the next deployment to go live in 1-2 days.

gbaz:
Yeah this gets a 👍 

Also, if its quick for you, you can add some wording pointing out the same to the signup confirmation email itself at: https://github.com/haskell/hackage-server/blob/master/datafiles/templates/UserSignupReset/SignupConfirmation.email.st

cocreature:
I’ve added the same note to the signup confirmation email so I think this is good to go.

PR #688, opened 2018-03-14
ux cleanup

### LINKED ISSUE #666: Streamline package page wording
From reddit:  "you can remove some of the static vocabulary that surrounds data and just present the data itself. Two cases where I'd do this are replacing The X Package with just X and Tags: tag1, tag2 with just #tag1 #tag2."

### LINKED ISSUE #686: Change revision wording on package pages
As per discussion on the ecosystem-requests uncurated ticket:

> If I'm right, I urge you to change "Updated" to "Revisions:". (And I'd re-order the info to "Revisions: Revision 1 made by Henning on 16 Feb 18".)
> I also wonder about putting Revisions: right up under Versions:. After all, a revision is really a kind of sub-version.

The idea would be to move "Updated" to under "Versions" and name it "Revised". Also, to indicate in the text somewhere if this was a non-maintainer revision.

gbaz:
Resolves #686  and resolves #666

### DIFF
--- a/datafiles/templates/Html/package-page.html.st
+++ b/datafiles/templates/Html/package-page.html.st
@@ -22,9 +22,9 @@
 
   <div id="content">
 
-    <h1>The $package.name$ package</h1>
+    <h1>$package.name$</h1>
     <div style="font-size: small">
-      [ <a href="/packages/tags">Tags: </a>$tags$ ]
+      [ $tags$ ]
       [ <a href="/package/$package.name$/tags/edit">Propose Tags </a> ]
     </div>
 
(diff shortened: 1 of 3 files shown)


datafiles/templates/Html/package-page.html.st
```diff
@@ -89,6 +89,13 @@
           <td>$package.maintainer$</td>
         </tr>
 
+        $if(hackage.hasUpdateTime)$
+        <tr>
+          <th>Revisions</th>
```
hvr: "Revisions" sounds odd to me; iirc I suggested somewhere on that long thread to use "Revised" instead (as it aligns better with the other label "Uploaded")

gbaz: sure, changed.

PR #690, opened 2018-03-15
Upgrade to Cabal-2.2

hvr:


Distribution/Server/Features/Core/Backup.hs
```diff
@@ -194,9 +194,12 @@ partialToFullPkg (pkgId, PartialPkg{..}) = do
...
 
     let (latestCabalFile, _) = last cabalRevisions
-    case parseGenericPackageDescription (cabalFileString latestCabalFile) of
-      ParseFailed err -> fail $ show (locatedErrorMsg err)
-      ParseOk _ _     -> return ()
+        filename = display pkgId ++ ".cabal"
+
+    case runParseResult $ parseGenericPackageDescription $
+         BS.toStrict $ cabalFileByteString latestCabalFile of
+      (_, Left (_, errs)) -> fail $ unlines (map (showPError filename) errs)
```
phadej: good idea to `unlines`!

Distribution/Server/Features/EditCabalFiles.hs
```diff
@@ -153,13 +152,10 @@ diffCabalRevisionsByteString oldRevision newRevision =
...
     parseSpecVerCheck = case parseGenericPackageDescriptionChecked newRevision of
-       (_,ParseFailed err) -> Just $ showError (locatedErrorMsg err)
-       (True, ParseOk {})  -> Nothing
-       (False,ParseOk {})  -> Just "The 'cabal-version' field could not be properly parsed"
-
-    showError (Nothing, msg) = msg
-    showError (Just n, msg) = "line " ++ show n ++ ": " ++ msg
-
+       (True, _, Right _)      -> Nothing -- parsing successful
+       (_, _, Left (_, err:_)) -> Just $ showPError "" err -- TODO: show all errors
```
phadej: `unlines` here too?

Distribution/Server/Features/Mirror.hs
```diff
@@ -239,14 +240,16 @@ mirrorFeature ServerEnv{serverBlobStore = store}
         fileContent <- expectTextPlain
         time <- liftIO getCurrentTime
         let uploadData = (time, uid)
-        case parseGenericPackageDescription . unpackUTF8 $ fileContent of
-            ParseFailed err -> badRequest (toResponse $ show (locatedErrorMsg err))
-            ParseOk _ pkg | pkgid /= packageId pkg ->
+            filename = display pkgid <.> "cabal"
+
+        case runParseResult $ parseGenericPackageDescription $ BS.L.toStrict $ fileContent of
+            (_, Left (_, err:_)) -> badRequest (toResponse $ showPError filename err)
```
phadej: as well as here

Distribution/Server/Features/Tags.hs
```diff
@@ -323,20 +322,22 @@ constructCategoryTags = map (tagify . map toLower) . fillMe . categorySplit . ca
...
-    in licenseToTag l
+    in [] -- licenseToTag l
     ++ (if hl then [Tag "library"] else [])
     ++ (if he then [Tag "program"] else [])
 -- These tags are too noisy and don't provide a good signal anymore
 --    ++ (if ht then [Tag "test"] else [])
 --    ++ (if hb then [Tag "benchmark"] else [])
     ++ constructCategoryTags desc
   where
+{- temporarily disabled until we figure out whether mapping SPDX to tags is desirable
```
phadej: we can map "simple identifiers" to tags, i.e .`License (ELicense (ELicenseId licId) Nothing)`

Distribution/Server/Packages/Unpack.hs
```diff
@@ -208,6 +209,18 @@ basicChecks pkgid tarIndex = do
   when (specVer < mkVersion [1,2]) $
     throwError "'cabal-version' must be at least 1.2"
 
+  -- Reject not well-defined cabal spec versions on upload; TODO:
+  -- factor out these version checks into a function
+  when (specVer >= mkVersion [1,25] && specVer < mkVersion [2]) $
+    throwError "'cabal-version' in unassigned >=1.25 && <2 range; use 'cabal-version: 2.0' instead"
+
+  when (specVer >= mkVersion [2,1] && specVer < mkVersion [2,2]) $
```
phadej: :+1:

Distribution/Server/Features/EditCabalFiles.hs
hvr: I didn't because I'm not sure it would render sensible; note that this may be converted into HTML.

phadej: :+1:

hvr:
I've merged this and will revisit the `unlines` issue when I have more time to figure out the HTML-rendering part

phadej:
note: errors might be multiline already today (`parsec`). We might need to hack around that. We can fix that for next `Cabal` though, if we know "how" :)

PR #693, opened 2018-03-16
Stylesheet revamp

gbaz:
Should supercede #648

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,9 @@
 [submodule "vendor/cabal"]
 	path = vendor/cabal
-	url = ../cabal
+	url = https://github.com/haskell/cabal.git
 	branch = 2.0
+
+[submodule "vendor/cabal-parsers"]
+	path = vendor/cabal-parsers
+	url = https://github.com/hvr/cabal-parsers.git
+	branch = master
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,10 @@ matrix:
     # env: TEST=--disable-tests BENCH=--disable-benchmarks
       addons: {apt: {packages: [ghc-ppa-tools,cabal-install-head,ghc-8.2.1], sources: [hvr-ghc]}}
 
+  allow_failures:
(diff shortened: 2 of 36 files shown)


gbaz:
example screenshot:
image

gbaz:
Added the two-col changes on top of this since they were very popular

image

gbaz:
(note: this version is responsive. when the window drops below 980, it falls back to one-column with the properties table following the description, as before)

gbaz:
Pushing a change to get the dependencies to align nicely: 
image

mikeplus64:
I might be alone here, but I much prefer the current colour scheme. The new fonts look good and layout changes are welcome, but to me, the orange links in the older theme stand out much better. This IMO degrades readability. Maybe I am just luddite, but I also just don't 'like' the choice of purple...

This is so clearly subjective that maybe it's a good idea to have the option in Hackage to switch between some "agreed upon" stylesheets? We used to be able to switch between the gloriously ugly :-) (and old) "Ocean" theme and the relatively new and fancy current one, which was pretty nice.

ocramz:
image

The dependencies don't seem to align properly

fredefox:
@ocramz As I understand it, this is the default behavior when multiple deps fit in one line.

gbaz:
This is by design. Each unit between each comma is nonbreaking now.

Profpatsch:
@gbaz For https://hackage.haskell.org/package/git-annex the right column kind of breaks.

For 125% it sometimes starts scrolling to the right: https://github.com/haskell/haddock/pull/721#issuecomment-374914421

gbaz:
What is the issue with git-annex? Just the sea of version numbers? 

For https://hackage.haskell.org/package/abeson the problem is the long name/email address with no whitespace to break on. It could be given the `word-wrap` style but then email addresses in general would break between any two characters, even when a proper whitespace break was possible, which also seems unfortunate, and would affect far more pages. Not sure about the right solution.

Issue #667, opened 2018-02-05
Collapsable metadata sections on package page


gbaz:
Metadata should be able to be collapsed en masse.

Also note, from reddit:  "Some of the meta information can also be summarised differently. For example, there could be a small numeric badge saying how many dependency which can then be clicked and expanded to the full glory. Similarly for older versions, it could be a badge saying (42 older versions). Even license information can be badged."

hvr:
I'm strongly against hiding/compacting dependency information by default, as that's one of the core things I look for when looking at a package on Hackage. Having to manually expand it would be a big regression for me.

gbaz:
Another common suggestion on the thread seems to have been multicolumn as per https://www.npmjs.com/package/twitter (or perhaps with the metadata on the lhs instead?)

gbaz:
resolved by the two-col layout imho.

PR #694, opened 2018-03-16
Two Column Layout for Package Pages [Do Not Merge]

gbaz:
This implements a two-column layout for package pages, pushing the description information to the right hand column, as suggested in #667 

It is independent of the broader redesign (#648) but I think would be merged after it, as complimentary, and so we can evaluate how it looks in the context of the redesign. Some screenshots follow:

image

image

image

Note that for lens, the image has to be resized or it causes problems. My hope is with the more percentage-based widths of the redesign we can avoid this -- otherwise a hack will be necessary.

### DIFF
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -579,7 +579,7 @@ ul.links li form button:hover {
 }
 
 #module-list li {
-  clear: right;
+/*  clear: right; */
 }
 
 #module-list span.collapser,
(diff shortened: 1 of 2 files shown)


ocramz:
Looks great! Perhaps the Dependencies could be shown one per row ?

robertmassaioli:
The dependencies are begging for a bullet list. I just wanted separate out that thought from the "Looks great" comment. Cheers.

gbaz:
I've now merged these changes over to the new layout: https://github.com/haskell/hackage-server/pull/693

gbaz:
closing in favor of the new layout PR

PR #695, opened 2018-03-17
wording about organizations on user signup form

gbaz:
Clarifies things to answer questions like: https://www.reddit.com/r/haskell/comments/7yj2rs/q_hackage_releases_for_orgs/

Also encourages FirstnameLastname style a bit more strongly.


datafiles/templates/UserSignupReset/SignupRequest.html.st
```diff
@@ -15,6 +15,8 @@ $hackagePageHeader()$
 In particular you need an account to be able to upload or help maintain packages.
 
 <p>Using the form below you can register an account.
+The FirstnameLastname format of usernames is encouraged but not required.
```
hvr: almost pointless to encourage something w/o any rationale and in the same sentence point out it's optional anyway... :-)

PR #698, opened 2018-03-18
add redirect on package.tar.gz download

### LINKED ISSUE #659: import CDN download count directly
We now have tarballs behind the CDN, but not html pages. This is pretty good, but we have the issue where the download counts are now wildly inaccurate again, because the CDN is in the middle. We can either import cdn stats, or we can configure the cdn to just give a little ping on each download -- e.g. for an ecache tag?

gbaz:
Resolves #659

The approach here is different than in that ticket. We just check if there is a "?real=true" at the end of the .tar.gz url. If there isn't, we bump the download count and redirect to a url that does have the path. If there is, then we just return the .tar.gz.

That way the cdn can cache the real url, and everything works as is.

### DIFF
--- a/Distribution/Server/Features/Core.hs
+++ b/Distribution/Server/Features/Core.hs
@@ -677,21 +677,28 @@ coreFeature ServerEnv{serverBlobStore = store} UserFeature{..}
       return . toResponse $ Array (Vec.fromList json)
 
     -- result: tarball or not-found error
+    -- note: this has a redirect gimmick so that we can cache the real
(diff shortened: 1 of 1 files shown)


gbaz:
This breaks some tests, they'll need to be fixed.

hvr:
Let's hold off on this one for a bit to get us more time to consider the consequences for Hackage clients out there...

gbaz:
Indeed, we're blocked on at least changes to the hackage-security backends, and perhaps other tools as well.

PR #714, opened 2018-03-20
move doc index link 

### LINKED ISSUE #400: A link to Index from the package page
It would be useful to have a link to the _Index_ from the main package page on the top right corner, i.e.

```
https://hackage.haskell.org/package/foobar-0.0.0
```

This makes it a little easier to find specific identifiers by reducing the number of mouse-clicks to _doc-index-All.html_ from 3 to 2.

gbaz:
Resolves #400

Under this change, the link will be right under `Modules` like this:

image

### DIFF
--- a/Distribution/Server/Pages/Package.hs
+++ b/Distribution/Server/Pages/Package.hs
@@ -32,6 +32,7 @@ import Distribution.Server.Packages.ModuleForest
 import Distribution.Server.Packages.Render
 import Distribution.Server.Users.Types (userStatus, userName, isActiveAccount)
 import Data.TarIndex (TarIndex)
+import qualified Data.TarIndex as Tar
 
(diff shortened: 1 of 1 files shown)


gbaz:
ugh, actually the "point to All" thing doesn't always work :-/

I guess haddock _conditionally_ creates an "All" based on how large the module is?

I'll remove that part of this PR. So what's the right way to not point to an empty page full of characters then? Perhaps haddock itself should be changed to split less aggressively? (And to make sure the index-doc page itself is always at least a little "interesting"?)

gbaz:
(if we want to get fancy, we can always probe the tarball for this, i guess...)

gbaz:
ok, now we do it the  right way and only provide doc-index-All if it exists.

PR #719, opened 2018-03-21
Fix floating overlap of sidebar

ClaireNeveu:
This PR fixes a few problems with the new package page layout. The main issue is shown in the following screenshots.

## Before
hackage-before

## After
hackage-after

Note that when the screen is small, the properties will appear after the main documentation. This is consistent with cargo, npm, and rubygems (I didn't check any other language's package repos).

P.S. I almost didn't contribute this change since the build instructions listed in the README did not work.

### DIFF
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -220,9 +220,7 @@ pre + pre {
 /* @group properties */
 
 div #properties {
-    float:right;
     min-width:350px;
-    background: #fefefe;
     width:25vw;
     margin-left:15px;
     margin-bottom:15px;
@@ -302,11 +300,16 @@ table.properties td, table.properties th {
     display: inline-table;
     margin: 2px 1em 0 2em;
   }
+
+  #content {
+    flex-direction: row;
+  }
 }
 
 @media only screen and (max-width: 950px) {
   #content {
     width: 88vw;
+    flex-direction: column;
   }
 
   #page-header {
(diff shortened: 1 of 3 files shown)


ClaireNeveu:
Considerable whitespace changes in this one, consider viewing with the diff with https://github.com/haskell/hackage-server/pull/719/files?w=1

gbaz:
Thanks for the PR! I think it captures one approach very well -- which is to just treat the right hand as a full column and not a floating div -- so nothing wraps around it, but instead it pushes everything to the left. This makes things more uniform, but at the cost of space efficiency. It is a good point in the design space to see if people like.

But even if people agree with the overall approach, there are a few other things I would like to raise.

I think it is wrong for the properties to move all the way to the bottom. While we want to make better use of space, we also do not want to hide them too much -- people find them important! (I have already seen some complaints about "pushing them off to the right" -- so I think people would find the bottom even more frustrating -- especially for packages with long readmes).

Also, now the properties box starts flush with the description, below the package name/title. This PR raises it to flush with the title, which I think is not as nice?

There are two other specific problems I found too, after testing on a few different packages with different content. One problem comes with "stratosphere" which has enormously long module names. The sidebar gets pushed all the way off the screen.

The other problem comes with "3dmodels" which has very _short_ content. In this case, the sidebar becomes flush with the short conent, and so is too far to the left.

gbaz:
I think the current solution we have should resolve this issue?

gbaz:
(i.e. after our latest redeploy)

PR #735, opened 2018-03-30
allow cc0 in older cabal files

gbaz:
cf haskell/cabal#5238

### DIFF
--- a/Distribution/Server/Packages/Unpack.hs
+++ b/Distribution/Server/Packages/Unpack.hs
@@ -38,6 +38,7 @@ import Distribution.Text
 import Distribution.Server.Util.ParseSpecVer
 import qualified Distribution.SPDX as SPDX
 import qualified Distribution.Compat.ReadP as Parse
+import qualified Distribution.License as OldLicense
 
 import Control.Monad.Except
          ( ExceptT, runExceptT, MonadError, throwError )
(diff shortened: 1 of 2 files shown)


phadej:
`cabal-version` is parsed always first.

hvr:
The thing I was worried was (and which Oleg confirmed to me), is that e.g. `version:1.12` + `license: CC0-1.0` won't be mapped to the proper SPDX type, and end up as `LicenseRef:CC0-1.0`   if we don't handle this hardwired exception in `lib:Cabal` proper.

phadej:
We could change licenseTo Spdx to try to parse UnknownLicense, PVP-wise it might be just "new feature", not "breaking change"

gbaz:
> The thing I was worried was (and which Oleg confirmed to me), is that e.g. version:1.12 + license: CC0-1.0 won't be mapped to the proper SPDX type, and end up as LicenseRef:CC0-1.0 if we don't handle this hardwired exception in lib:Cabal proper.

@hrv what would be the consequences of this for hackage? would it just mess up other elements of the UI/display?

hvr:
@gbaz I'm not sure; I'm generally worried that having `licence: CC0-1.0` map to two different things (rather than the unique exact meaning) may result in weirdness down the road. I'll trust @phadej judgement whether it's a problem for lib:Cabal to have these "degenerate licence levels"... :-)

gbaz:
superseded by: https://github.com/haskell/hackage-server/commit/af9096e00bf251c6919e568b0d12441e6715cb65

PR #736, opened 2018-04-01
Fix search bugs

gbaz:
This fixes two bugs in search -- first by not concating component strings of markdown, but instead joining them with whitespace (pre-tokenization). Second a stupid order-of-arguments error that meant the extra stemming added in https://github.com/haskell/hackage-server/pull/633 wasn't working properly. This should go a fair way to addressing https://github.com/haskell/hackage-server/issues/615


hvr:
@gbaz does this address that base64 issue I mentioned?

PR #743, opened 2018-04-13
fix faviconURL

mustafa-cosar:


phadej:
Merged as a part of #744, thanks!

PR #755, opened 2018-05-11
uploader group add process improvements

gbaz:
This does three things. First, adds trustees to those allowed to add uploaders. Second, moves the add bit to the _top_ rather than bottom of the edit group html, above rather than below existing group members. Third, gives a more informative error when uploads fail due to users not being in the uploaders group.


gbaz:
After this is merged we should also modify the central-server branch to point to trustees as the contact point for requests rather than admins.

PR #776, opened 2018-07-10
Add blockquote styling

### LINKED ISSUE #775: Block quotes end up in the html but aren't styled
This leads to some confusion on pages like [this one](https://hackage.haskell.org/package/dhall-to-cabal).
The paragraph that starts with "a system for building..." should actually be a block quote and it is, it just doesn't look like one.

toonn:
Fixes #775

I guessed at a border color that fit with the purple theme and selected a lighter tint for the background.
As you can probably see from the padding rule I'm not very familiar with css, I'm not sure why the top padding was so hard to change but it was always much larger than the bottom padding. I also took a guess at the right padding so as to make it symmetrical with the left padding.

Feel free to suggest better colors or padding.

### DIFF
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -209,6 +209,13 @@ pre + pre {
   margin-top: 0.5em;
 }
 
+blockquote {
+  border-left: 3px solid #c7a5d3;
+  background: #eee4f1;
+  margin: 0.5em;
+  padding: 0.0005em 0.3em 0.5em 0.5em;
+}
+
 .src {
   background: #f4f4f4;
   padding: 0.2em 0.5em;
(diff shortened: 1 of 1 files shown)


phadej:
Screenshot would be nice :)

toonn:
Let there be screenshots!
No styling:
screenshot_2018-07-10 dhall-to-cabal compile dhall expressions to cabal files

My guess at a tint:
screenshot_2018-07-10 dhall-to-cabal compile dhall expressions to cabal files 1

Same background color as pre:
screenshot_2018-07-10 dhall-to-cabal compile dhall expressions to cabal files 2

toonn:
@phadej, @gbaz, just pinging. Is there anything more you want me to change about this?

PR #777, opened 2018-07-14
Enable new syntax highlighter and quick jump files

glguy:


glguy:
closing in favor of a change that only add quickjump (new highlighter is already being used)

PR #778, opened 2018-07-14
Generate quickjump index file in doc builder

glguy:


harpocrates:
I'm not sure which version of cabal-install it ended up in, but there is a [`--haddock-quickjump` option](https://github.com/haskell/cabal/blob/2ce4858881924fbd1ed22e133b6e7a999737b965/Cabal/Distribution/Simple/Setup.hs#L1525-L1528) too.

glguy:
@harpocrates That's good to know. It doesn't seem to be in the release version given this test. I don't know what the builder is using, however.

```
$ cabal install --haddock-quickjump
cabal: unrecognized 'install' option `--haddock-quickjump'
$ cabal --version
cabal-install version 2.2.0.0
compiled using version 2.2.0.0 of the Cabal library 
```

gbaz:
I merged, rebuilt on the docbuilder server, and think I have the right exe in production. We should confirm on the next build to be sure.

gbaz:
works!

glguy:
Thanks for handling this!

PR #801, opened 2019-01-07
Run Haddock on package titles

harpocrates:
See http://hackage.haskell.org/package/ghc-boot-th for an example of where this
matters.

### DIFF
--- a/Distribution/Server/Pages/Package.hs
+++ b/Distribution/Server/Pages/Package.hs
@@ -103,7 +103,8 @@ packagePage render headLinks top sections
     bodyTitle = case synopsis (rendOther render) of
       ""    -> h1 << pkgName
       short -> h1 << [ toHtml (pkgName ++ ": ")
-                     , small (toHtml short)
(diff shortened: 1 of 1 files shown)


hvr:
@harpocrates we can't do that; `title` and `synopsis` are intentionally *not* supposed to have full haddock markup; instead I was planning to support a *very* limited markup subset.

harpocrates:
> @harpocrates we can't do that; `title` and `synopsis` are intentionally _not_ supposed to have full haddock markup; instead I was planning to support a _very_ limited markup subset.

That shouldn't be an issue, we'll just another `limitedHtmlMarkup :: DocMarkup`. What is the limitation (so I know what we can let through)?

harpocrates:
Closing for now, since:

  1. it isn't clear that we want to support arbitrary markup in titles (or even _what_ markup we want to support)
  2. the `synopsis` field in .cabal files isn't specified as being Haddock formatted

PR #813, opened 2019-03-10
Tasty update

phadej:
This is mechanical PR, I changed from `test-framework` to `tasty` as latter was used in some other test already.


phadej:
ping

PR #814, opened 2019-03-11
Use zlib-0.6

phadej:
Note that code path in `RequestContentTypes.hs` is only when `Request-Encoding: gzip`, which is usually **not** when we upload `tar.gz` files!

On package upload we use `decompressNamed` which is a wrapper around `decompress`; the code path, which works the same.


phadej:
ping

PR #820, opened 2019-03-29
Fix MemSize instance for ShortByteString

intricate:
`ShortByteString` actually has 4 words overhead + words consumed by the
`ByteArray#` payload (word-aligned).


intricate:
bump :)

PR #826, opened 2019-04-26
fix postfix apt installation & remove ** usage so cabal copy works.

dsturnbull:
Minimum changes I needed to have `docker build .` work.

* `ghc-8.2.1` is no longer available in `ppa:hvr/ghc`
* The `**` glob didn't work so I flattened the `data-files` section
* `cabal-install-2.4` was also needed
* `postfix` refuses to obey `-y`, but a noninteractive frontend for `apt-get` works.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ RUN apt-get update
 RUN DEBIAN_FRONTEND=noninteractive apt-get install -yy unzip libicu-dev postfix
 RUN apt-get install -y ghc-8.2.2 cabal-install-2.4
 ENV PATH /opt/ghc/bin:$PATH
-RUN cabal update
+RUN cabal v2-update
 
 # Required Header files
 RUN apt-get install -y zlib1g-dev libssl-dev
@@ -35,13 +35,11 @@ RUN apt-get install -y zlib1g-dev libssl-dev
 RUN mkdir /build
 WORKDIR /build
 ADD ./hackage-server.cabal ./hackage-server.cabal
-RUN cabal sandbox init
-# TODO: Switch to Nix-style cabal new-install
-RUN cabal install --only-dependencies --enable-tests -j --force-reinstalls
-ENV PATH /build/.cabal-sandbox/bin:$PATH
+RUN cabal v2-install -j --force-reinstalls
+ENV PATH /root/.cabal/bin:$PATH
 
 # needed for creating TUF keys
-RUN cabal install hackage-repo-tool
+RUN cabal v2-install hackage-repo-tool
 
 # add code
 # note: this must come after installing the dependencies, such that
@@ -56,13 +54,13 @@ RUN hackage-repo-tool create-root --keys keys -o datafiles/TUF/root.json
 RUN hackage-repo-tool create-mirrors --keys keys -o datafiles/TUF/mirrors.json
 
 # build & test & install hackage
-RUN cabal configure -f-build-hackage-mirror --enable-tests
-RUN cabal build
+RUN cabal v2-configure -f-build-hackage-mirror --enable-tests
+RUN cabal v2-build
 # tests currently don't pass: the hackage-security work introduced some
 # backup/restore errors (though they look harmless)
 # see https://github.com/haskell/hackage-server/issues/425
 #RUN cabal test
-RUN cabal copy && cabal register
+RUN cabal v2-install all
 
 # setup server runtime environment
 RUN mkdir /runtime
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,8 @@ RUN apt-add-repository ppa:hvr/ghc
 RUN apt-get update
 
 # Dependencies
-RUN apt-get install -yy unzip libicu-dev postfix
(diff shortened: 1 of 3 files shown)


hvriedel:
> The ** glob didn't work so I flattened the data-files section

That's news to me; can you elaborate on why you think those wouldn't work?

hvr:
> The ** glob didn't work so I flattened the data-files section

That's news to me; can you elaborate on why you think those wouldn't work?

And btw, the Dockerfile desperately needs to be converted to use v2-* commands, as building with the old-style commands is not something well-tested nor actively supported.

dsturnbull:
> > The ** glob didn't work so I flattened the data-files section
> 
> That's news to me; can you elaborate on why you think those wouldn't work?

I can probably fix this up. I think I was using an older cabal version initially.

> 
> And btw, the Dockerfile desperately needs to be converted to use v2-* commands, as building with the old-style commands is not something well-tested nor actively supported.

This one is more problematic. The sandbox command is v1-only. Not sure if I can just skip it or not.

hvr:
> This one is more problematic. The sandbox command is v1-only. Not sure if I can just skip it or not.

Well, there's a todo-comment near the `sandbox` command which hints at it that `cabal sandbox` is deprecated and no longer needed in the v2-build paradigm (i.e. v2-build was created to properly solve the problem that `sandbox` workarounded...) :-)

https://github.com/haskell/hackage-server/blob/b4fdd6768d145d316d44d0bc855e8bf0e891fcec/Dockerfile#L38-L40

dsturnbull:
The `**` syntax just doesn't work with the build as it is.

```
/root/.cabal/store/ghc-8.2.2/incoming/new-7/root/.cabal/store/ghc-8.2.2/hackage-server-0.5.1-3eca8c7e97a2dae3ce64cbec3e37887041b4d7e1feba6279abc894a42c624ca2/share/templates/Html: copyFile: does not exist (No such file or directory)
```

There is no reason why cabal-2.4 can't figure it out, but it doesn't.

I've pushed the v2- style updates though.

peti:
I have run a build of the current `master` branch with cabal-install version 2.4.1.0 and ghc 8.2.2 on NixOS, and I get the same error as the docker build above (which uses the same setup on Ubuntu):

~~~
$ cabal new-install -j .
Wrote tarball sdist to /home/simons/src/hackage-server/dist-newstyle/sdist/hackage-server-0.5.1.tar.gz
Resolving dependencies...
Build profile: -w ghc-8.2.2 -O1
In order, the following will be built (use -v for more details):
 - hackage-server-0.5.1 (lib:lib-server) (requires build)
 - hackage-server-0.5.1 (exe:hackage-build) (requires build)
 - hackage-server-0.5.1 (exe:hackage-server) (requires build)
 - hackage-server-0.5.1 (exe:hackage-mirror) (requires build)
[...]
[169 of 169] Compiling Distribution.Server ( Distribution/Server.hs, dist/build/lib-server/Distribution/Server.o )
/home/simons/.cabal/store/ghc-8.2.2/incoming/new-8190/home/simons/.cabal/store/ghc-8.2.2/hackage-server-0.5.1-445eec1b86c4b1649e7cb92b3c4c62208ccc4f9744ba085d51bd7b47dca66816/share/templates/Users: copyFile: does not exist (No such file or directory)
cabal: Failed to build lib:lib-server from hackage-server-0.5.1 (which is
required by exe:hackage-build from hackage-server-0.5.1, exe:hackage-server
from hackage-server-0.5.1 and others). See the build log above for details.
~~~

Not that the error occurs while `cabal` is still *compiling*. If I run `new-build` rather then `new-install`, then the build succeeds.

peti:
For what it's worth, https://cloud.docker.com/repository/registry-1.docker.io/psimons/hackage-server/builds/fb8b0316-6639-41d4-a91f-e2136e2eb212 shows a log of the complete docker build (and the failure):

~~~
[...]
[169 of 169] Compiling Distribution.Server ( Distribution/Server.hs, dist/build/lib-server/Distribution/Server.o )
/root/.cabal/store/ghc-8.2.2/incoming/new-6/root/.cabal/store/ghc-8.2.2/hackage-server-0.5.1-7fefae90d236fca2cb7bc003f1543ad88119f0e9886624c2a5520a059b0688a0/share/templates/Users: copyFile: does not exist (No such file or directory)
Warning: Some package(s) failed to build. Try rerunning with -j1 if you can't
see the error.
cabal: Failed to build lib:lib-server from hackage-server-0.5.1 (which is
required by exe:hackage-build from hackage-server-0.5.1, exe:hackage-server
from hackage-server-0.5.1 and others). The failure occurred during the final
install step.
~~~

quasicomputational:
Ah, this is getting bitten by haskell/cabal#6125 for `data-files` / `copyFile` thing. That's fixed in Cabal HEAD.

peti:
https://github.com/haskell/hackage-server/pull/836 fixes the build. It uses a locally compiled version of `cabal-install` from the current git master.

ocramz:
@peti shall we close this?

peti:
> @peti shall we close this?

Yes, I all issue this PR addresses are now fixed in `master`.

PR #827, opened 2019-04-30
"cabal-version: 2.0" fails HighLevelTest [WIP]

dsturnbull:
I'm trying to figure out why hackage-server doesn't accept libraries with cabal-version >= 2.0. This test captures that problem.


hvr:
That's no big mystery; it's simply that the test needs to be modified to make sense for cabal-version >= 2 as those don't show up in the legacy index

gbaz:
I think this is outdated and we accept newer lib versions now.

Issue #835, opened 2019-07-11
Dockerfile does not build


peti:
While building the docker container,  the installation of `postfix` requires user interaction that does not work:

~~~
debconf: unable to initialize frontend: Dialog
debconf: (TERM is not set, so the dialog frontend is not usable.)
debconf: falling back to frontend: Readline
Postfix Configuration
---------------------

Please select the mail server configuration type that best meets your needs.

 No configuration:
  Should be chosen to leave the current configuration unchanged.
 Internet site:
  Mail is sent and received directly using SMTP.
 Internet with smarthost:
  Mail is received directly using SMTP or by running a utility such
  as fetchmail. Outgoing mail is sent using a smarthost.
 Satellite system:
  All mail is sent to another machine, called a 'smarthost', for delivery.
 Local only:
  The only delivered mail is the mail for local users. There is no network.

  1. No configuration  3. Internet with smarthost  5. Local only
  2. Internet Site     4. Satellite system
General type of mail configuration: 
~~~

There is no way to answer that question; user input in the shell has no effect.

peti:
https://github.com/haskell/hackage-server/pull/826 addresses this issue, apparently, but the PR seems to be stuck.

hvr:
It's fixed properly now via 3a5bc1a4b8e519cbf21bbb3f9f53a064bab42b37

PR #836, opened 2019-07-12
Fix the Dockerfile so that it actually builds

peti:
Fixes https://github.com/haskell/hackage-server/issues/835.
Supersedes: https://github.com/haskell/hackage-server/pull/826.


peti:
Ping?

ocramz:
@peti Thank you!

hvr:
hrm... I'm not happy with the Dockerfile as is, specifically the part where it builds from Git; But since this got merged already in the current I'll have to follow-up with a commit to master later today

PR #845, opened 2019-09-23
fix build with ghc 8.6.5

peti:
No older compilers were harmed by those changes: https://travis-ci.org/haskell/hackage-server/builds/588445641

### DIFF
--- a/Distribution/Server/Features/Security/State.hs
+++ b/Distribution/Server/Features/Security/State.hs
@@ -267,13 +267,15 @@ updateSecurityState maxAge newUpdate = do
     when (needUpdate oldUpdate newUpdate) $ do
       setTufUpdate newUpdate
       updateSnapshotAndTimestamp (tufUpdateTime newUpdate)
-    SecurityState {
-      securityStateFiles  = Just SecurityStateFiles {
-                              securitySnapshot  = snapshot,
-                              securityTimestamp = timestamp
-                            }
-    } <- State.get
-    return (timestamp, snapshot)
+    st <- State.get
+    case st of
+      SecurityState {
+        securityStateFiles  = Just SecurityStateFiles {
+                                securitySnapshot  = snapshot,
+                                securityTimestamp = timestamp
+                              }
+      } -> return (timestamp, snapshot)
+      _ -> error "updateSecurityState: unexpected state"
   where
     needUpdate :: Maybe TUFUpdate -> TUFUpdate -> Bool
     needUpdate Nothing    _   = True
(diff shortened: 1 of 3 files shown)


Distribution/Server/Features/Security/State.hs
```diff
@@ -267,13 +267,15 @@ updateSecurityState maxAge newUpdate = do
...
-    return (timestamp, snapshot)
+    st <- State.get
+    case st of
+      SecurityState {
+        securityStateFiles  = Just SecurityStateFiles {
+                                securitySnapshot  = snapshot,
+                                securityTimestamp = timestamp
+                              }
+      } -> return (timestamp, snapshot)
+      _ -> error "updateSecurityState: unexpected state"
```
hvr: I'm not sure this is the proper fix; what `Monad` are we in here? can it express failure in a more exact way than by a "pure" `error`?

hvr:
> No older compilers were harmed 

:laughing:

Distribution/Server/Features/Security/State.hs
peti: That code runs in `acid-state`'s `Update` monad: http://hackage.haskell.org/package/acid-state-0.15.1/docs/Data-Acid.html#t:Update. It's not an instance of `MonadFail`. There is a legacy `fail` method, of course, but it's semantics are not clear to me. I'm not terribly familiar with that code, which is why I opted for the solution that I knew for sure would behave the same way the original version of the code does. I'd rather not screw up security critical code.

hvr: So, 

```hs
newtype Update st a = Update { unUpdate :: State st a }
    deriving (Monad, Functor, MonadState st)
```

inherits the `Monad` instance of `StateT _ Identity` which does indeed not have a good `fail` operation to offer... so I guess pure exceptions such as `error` are as good as it gets...

Distribution/Server/Features/Security/State.hs
```diff
@@ -292,22 +294,24 @@ updateSecurityState maxAge newUpdate = do
...
+          securityTarGzFileInfo = tarGzFileInfo,
+          securityTarFileInfo   = tarFileInfo
+        } -> return $ Just TUFUpdate {
+               tufUpdateInfoRoot    = fauxFileInfo root,
+               tufUpdateInfoMirrors = fauxFileInfo mirrors,
+               tufUpdateInfoTarGz   = tarGzFileInfo,
+               tufUpdateInfoTar     = tarFileInfo,
+               tufUpdateTime        = timestampTime
+             }
+        _ -> error "getTufUpdate: unexpted state"
```
hvr:
```suggestion
        _ -> error "getTufUpdate: unexpected state"
```

peti:
New test build results are at https://travis-ci.org/peti/hackage-server/builds/588835641 -- now including ghc 8.6.5.

hvr:
@peti thanks!

PR #848, opened 2019-10-05
Use ISO8601 based format for listing of recent additions

ad-si:
This will print
```
2019-10-05 09:18:07Z   barrucadu   irc-client-1.1.1.1
```
instead of
```
Sat Oct 5 09:18:07 UTC 2019   barrucadu   irc-client-1.1.1.1
```
on https://hackage.haskell.org/packages/recent

Since Hackage is an international project, it should also use the international date format as described in [ISO8601](https://en.wikipedia.org/wiki/ISO_8601).

### DIFF
--- a/Distribution/Client/Mirror/Repo/Hackage2.hs
+++ b/Distribution/Client/Mirror/Repo/Hackage2.hs
@@ -113,7 +113,7 @@ uploadPackage targetRepoURI' doMirrorUploaders pkginfo locCab locTgz = do
     tgzURI  = baseURI <//> display pkgid               <.> "tar.gz"
 
     putPackageUploadTime time = do
-      let timeStr = formatTime defaultTimeLocale "%c" time
+      let timeStr = formatTime defaultTimeLocale "%Y-%m-%d %H:%M:%S%EZ" time
       requestPUT (baseURI <//> "upload-time") "text/plain" (packUTF8 timeStr)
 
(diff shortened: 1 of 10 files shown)


hvr:
will this also take care of https://hackage.haskell.org/packages/recent/revisions.html ?

ad-si:
Good point @hvr. I  rechecked the code for other occurrences of the old datetime formatting and fixed them.

ad-si:
Also I changed the formatting a little to be compatible with RFC 3339.
I.e. `2019-10-05 09:18:07Z` instead of `2019-10-05 09:18:07 UTC`

phadej:
I'll merge this after #848, yet I'll change space to be `T` - even it looks a bit more human unfriendly.

ad-si:
Why? According to ISO:
> ISO 8601:2004(E)". ISO. 2004-12-01. 4.3.2 NOTE: By mutual agreement of the partners in information interchange, the character [T] may be omitted in applications where there is no risk of confusing a date and time of day representation with others defined in this International Standard.

Like you say, it will only make it less readable …

ad-si:
Oh, `omit` vs `replace`. My bad. Anyways, I think RFC 3339 is good enough. So I still would leave it off.

phadej:
Merged in https://github.com/haskell/hackage-server/pull/866. Thanks!

PR #853, opened 2019-11-06
README: Update macOS instructions for icu4c

dwijnand:
Running `brew link icu4c --force` resulted in:

    Warning: Refusing to link macOS-provided software: icu4c
    If you need to have icu4c first in your PATH run:
      echo 'export PATH="/usr/local/opt/icu4c/bin:$PATH"' >> ~/.bash_profile
      echo 'export PATH="/usr/local/opt/icu4c/sbin:$PATH"' >> ~/.bash_profile

    For compilers to find icu4c you may need to set:
      export LDFLAGS="-L/usr/local/opt/icu4c/lib"
      export CPPFLAGS="-I/usr/local/opt/icu4c/include"

    For pkg-config to find icu4c you may need to set:
      export PKG_CONFIG_PATH="/usr/local/opt/icu4c/lib/pkgconfig"

and doing that didn't allow hackage-server to build (`cabal build`).

Using `cabal v2-build --extra-lib-dirs=/usr/local/opt/icu4c/lib/
--extra-include-dirs=/usr/local/opt/icu4c/include` also didn't work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,10 +13,15 @@ uses the [icu4c](http://icu-project.org/apiref/icu4c/) library to build.
 
 You'll need to do the following to get hackage-server's dependency `text-icu` to build:
 
-### Mac OS X
+### macOS
 
     brew install icu4c
-    brew link icu4c --force
+
+Then create a `cabal.project.local` with
+
+    package text-icu
+      extra-include-dirs: /usr/local/opt/icu4c/include
+      extra-lib-dirs: /usr/local/opt/icu4c/lib
 
 ### Ubuntu/Debian
 


README.md
```diff
@@ -13,10 +13,15 @@ uses the [icu4c](http://icu-project.org/apiref/icu4c/) library to build.
 
 You'll need to do the following to get hackage-server's dependency `text-icu` to build:
 
-### Mac OS X
+### macOS
 
     brew install icu4c
-    brew link icu4c --force
+
```
hvr: I'd still briefly mention `brew link icu4c --force` as it might work on older macOS versions, and more importantly, it dumps you the paths you might need to add yourself in the `cabal.project.local`

So maybe something along the lines of

> In recent versions of macOS `brew link icu4c --force` will refuse to operate and suggest to set the include paths yourself; in such case create a `cabal.project.local` file like so

?

dwijnand: It might still work, but I think it was a bad idea anyway (thus the need for `--force`).

In terms of the path, those are stable, known paths.  Indeed some packages put those straight in their .cabal, [including test-icu](https://github.com/bos/text-icu/blob/5ab84d8106c06fbabb87a0f173faa6049775a148/text-icu.cabal#L84-L86) - if only that change were released to hackage...

Oh and when you `brew install icu4c` brew prints out its caveat text anyways:

```
==> Caveats
icu4c is keg-only, which means it was not symlinked into /usr/local,
because macOS provides libicucore.dylib (but nothing else).

If you need to have icu4c first in your PATH run:
  echo 'export PATH="/usr/local/opt/icu4c/bin:$PATH"' >> ~/.bash_profile
  echo 'export PATH="/usr/local/opt/icu4c/sbin:$PATH"' >> ~/.bash_profile

For compilers to find icu4c you may need to set:
  export LDFLAGS="-L/usr/local/opt/icu4c/lib"
  export CPPFLAGS="-I/usr/local/opt/icu4c/include"

For pkg-config to find icu4c you may need to set:
  export PKG_CONFIG_PATH="/usr/local/opt/icu4c/lib/pkgconfig"
```

dwijnand: Convinced, @hvr?

PR #865, opened 2020-03-22
Show ISO8601 and SHA256 of revisions

phadej:
The rendered table is ugly,
but this information is needed
(and extracting SHA256 is tricky).

Screenshot from 2020-03-22 16-47-14


phadej:
Another option is to render as 
Screenshot from 2020-03-22 16-53-11

There are no space for different time format, but it is not strictly required...

PR #866, opened 2020-03-22
ISO8601 time

phadej:
Use ISO8601 time. In many places the tooltip prints %c format, so if you want to know the day of the week, it is there :)

Screenshot from 2020-03-22 21-18-52
Screenshot from 2020-03-22 21-18-42

Supersedes #848

### DIFF
--- a/Distribution/Server/Framework/Templating.hs
+++ b/Distribution/Server/Framework/Templating.hs
@@ -24,6 +24,7 @@ module Distribution.Server.Framework.Templating (
     TemplateVal,
     templateDict,
     templateVal,
+    utcTimeTemplateVal,
     templateEnumDesriptor,
     templateUnescaped,
     ToSElem(..),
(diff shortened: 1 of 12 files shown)


Distribution/Server/Framework/Templating.hs
```diff
@@ -119,6 +124,18 @@ instance ToSElem PackageName       where toSElem = toSElem . display
 instance ToSElem Version           where toSElem = toSElem . display
 instance ToSElem PackageIdentifier where toSElem = toSElem . display
 
+-- | Format 'UTCTime' value as HTML.
+utcTimeTemplateVal :: String -> UTCTime -> (String, TemplateVal)
+utcTimeTemplateVal k utime = (k, TemplateVal (SBLE (stFromString utcSpanEl)))
```
phadej: This is not the cleanest way. Preferably we should have the `XHtml` based helper in one place, and use it here.

I'm not familiar enough with code base to know where such utils belong.

PR #871, opened 2020-05-16
Add package RSS feed

### LINKED ISSUE #203: RSS feeds for individual packages.
A wish: RSS/Atom feeds for each of the packages individually 

I think it might be useful to subscribe to the packages you are actively using as libraries or whatnot

geyaeb:
I maintain a couple of Linux packages of Haskell programs, having an RSS feed with new versions of packages on Hackage would be very useful.

This PR adds a new type of address of form `/package/name.rss` which generates an RSS feed for given package. At the moment, the feed consists of only uploaded versions but I can see possibility to add other package-specific events in the future.

Details:

* Link to RSS feed is next to "Versions" header on package's page
* Each feed item contains link to package's homepage (not on Hackage), author/uploader/maintainer and includes complete changelog, if exists
* Destination of feed item is package's Hackage page

This PR will resolve #203 .

### DIFF
--- a/Distribution/Server/Features.hs
+++ b/Distribution/Server/Features.hs
@@ -46,6 +46,7 @@ import Distribution.Server.Features.AdminLog            (initAdminLogFeature)
 import Distribution.Server.Features.HoogleData          (initHoogleDataFeature)
 import Distribution.Server.Features.Votes               (initVotesFeature)
 import Distribution.Server.Features.Sitemap             (initSitemapFeature)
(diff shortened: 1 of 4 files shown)


gbaz:
👍

gbaz:
Can you add revisions and maintainer group changes and build reports being added? Then this plus people setting up an "if this then that" or something could get us email notifications for maintainers.

geyaeb:
Good ideas. I will look into it…

PR #885, opened 2020-06-25
Converte candidate page to use templating engine and auto-delete published candidates

sitaochen:
- Converted individual candidate page and candidate index page to template-based generation like `servePackagePage` does
- Added two html templates, `candidate-page.html.st` and `candidate-index.html.st`, for these pages
- Delete candidates automatically after they are published by setting `doDelete` to `True`


sitaochen:
this is included in another pull request

PR #930, opened 2021-05-24
prelim fixes 810

gbaz:
This changes the bounds so hackage-server builds with ghc 8.10, and fixes some various small type errors that cropped up due to the migration. Arguably we should bump more bounds in a second pass to get in better shape still, but its a start!


hackage-server.cabal
```diff
@@ -349,8 +350,8 @@ library lib-server
   build-depends:
     , HStringTemplate       ^>= 0.8
     , HTTP                  ^>= 4000.3.6
-    , QuickCheck            ^>= 2.9
-    , acid-state            ^>= 0.14
+    , QuickCheck            > 2.9 && < 2.15
```
phadej, line 353: The `>` explicitly excludes version `2.9`. I suspect all lower bounds should been `>=`?

Yet, I'd suggest supporting only single major version, i.e. in this case `^>=2.14` (or even `^>=2.14.2` to go for the latest). Older and in between versions are probably never tested anyway.

gbaz:
testing reveals that this seems to work, so we'll merge to go with this for a deply. we'll do a fixup commit to actually try to get all the versions updated and tight soon.

gbaz:
(note our ci is busted on this because it only goes thru 8.8. we need to set up with a new ci anyway, as travis-ci.org is going away real soon now)

PR #936, opened 2021-05-29
Fix #866 which changed the datetime format to non iso8601

### LINKED ISSUE #933: Hackage revisions endpoint no longer using iso 8601 for `time`
Hi,

Not sure if this is an actual issue, but I thought I would raise it just in case. Prior to the hackage maintenance the revisions endpoint would return the following JSON.

```json
[
 {
    "user": "phadej",
    "number": 0,
    "time": "2020-07-14T18:40:29Z"
  }
]
```

The `time` is in iso 8601 format. However, after the maintenance it now is ` "time": "2020-07-14T18:40:29UTC"`, which is not iso 8601. This actually fails to decode with the default `aeson` decoder for `UTCTime`. As such I am using `parseTimeM False defaultTimeLocale "%Y-%-m-%-dT%H:%M:%SUTC"`.

You can reproduce with `curl -H "Accept: application/json" https://hackage.haskell.org/package/these/revisions/`

I think this should probably still be iso 8601 format?

AlistairB:
Closes #933

Specifically, `%EZ` will print the timezone name ie. `UTC`, when in iso8601 the UTC timezone is represented as `Z` indicating zero offset.

The other option is to use https://hackage.haskell.org/package/time-1.11.1.2/docs/Data-Time-Format-ISO8601.html#v:iso8601Show , however I wasn't sure if there might be other implications around using it ie. `defaulteTimeLocale` is no longer explicitly being specified.

cc @gbaz @phadej

### DIFF
--- a/Distribution/Client/Mirror/Repo/Hackage2.hs
+++ b/Distribution/Client/Mirror/Repo/Hackage2.hs
@@ -113,7 +113,7 @@ uploadPackage targetRepoURI' doMirrorUploaders pkginfo locCab locTgz = do
     tgzURI  = baseURI <//> display pkgid               <.> "tar.gz"
 
     putPackageUploadTime time = do
-      let timeStr = formatTime defaultTimeLocale "%Y-%m-%dT%H:%M:%S%EZ" time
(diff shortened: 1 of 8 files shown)


gbaz:
thanks!

PR #946, opened 2021-06-22
break before dots in module names

noughtmare:
See #928.

Now the module list of `prettyprinter` looks like this:

image


phadej:
This doesn't look very nice. I don't see a value of two columns on such narrow screens.

noughtmare:
The other options are making the threshold for the switch to one-column mode lower, but then people can just write even longer module names which will wrap again. Or we can use a flexbox as I experimented with in [this comment](https://github.com/haskell/hackage-server/issues/928#issuecomment-865891120).

phadej:
Let's discuss in #928.

PR #951, opened 2021-07-15
Handle HTTP 451 error code in mirror client

edsko:
There are some packages for which Hackage returns a 451 "Unavailable for Legal Reasons". At the moment the mirror client chokes on these and exits; we should treat this the same as a 404.


gbaz (approved):

gbaz:
Thanks!

PR #952, opened 2021-07-16
New mirror client flag: skip existing packages

edsko:
When mirroring to a local repo, using the `--skip-exists` command line flag will now check if the local file for a package already exists, and if so, skip downloading this package. It does not work for remote repos nor does it check checksums or hashes or anything like that. Nonetheless, I am finding it useful, as I was having trouble creating a full mirror of hackage using the mirror client; it would stop for some reason or other somewhere halfway, not have a local index yet, and start over entirely.

Probably the mirror client needs some love, but until somebody has time for that, this is probably an improvement. Feel free to ignore this PR of course if you feel it's not the right thing to do; just sharing in case it's useful.

### DIFF
--- a/Distribution/Client/Mirror/CmdLine.hs
+++ b/Distribution/Client/Mirror/CmdLine.hs
@@ -24,7 +24,8 @@ data MirrorOpts = MirrorOpts {
                     selectedPkgs    :: [PackageId],
                     continuous      :: Maybe Int, -- if so, interval in minutes
                     mo_keepGoing    :: Bool,
-                    mirrorUploaders :: Bool
+                    mirrorUploaders :: Bool,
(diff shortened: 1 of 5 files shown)


gbaz:
I agree that the whole mirror client approach should be done better imho (driven by the index-01.tar more directly rather than how it works now) -- basically because the current approach is terribly slow. I've only really used mirror client to set up a small local test server or to mirror a subuniverse of packages, not the whole thing. But that said, within the parameters of what exists, any improvements are always more than welcome, and this is a very helpful one. Thanks!

edsko:
Thanks :) Well, the good news is that between #951 and #952 I can now do a full local mirror of Hackage :)

PR #958, opened 2021-08-02
Hide documentation banner when package only contains an executable

### LINKED ISSUE #945: Banner "documentation unavailable" displayed for executables
If you look e.g. at https://hackage.haskell.org/package/alex , it flags in red "Documentation Unavailable".  However, this is an executable-only package which does not ship with haddockumentation.  It is probably better to not put the red banner up in such cases, as it suggests that something is missing/does not work when in reality everything is fine.

hsek:
Closes #945


gbaz:
Thanks!

PR #962, opened 2021-08-08
Add OpenGraph metadata for package pages

### LINKED ISSUE #912: Open Graph description is outdated on package pages
The `og:description` meta tag recommends using `cabal install package-name` for installing packages.
https://github.com/haskell/hackage-server/blob/4039daa37203c1db08c7053ae2751bcf24ae0674/datafiles/templates/Html/package-page.html.st#L15-L16
As of Cabal 3.0 with the new v2-* commands being default this is no longer the recommended way to install packages. It might confuse new Haskell users who see Discord embeds still recommending to use `cabal install` to install dependencies, a matter that already trips them up.
I'm not sure what should be put here instead, suggestions?

MorrowM:
Fixes #912, and adds the functionality to `master`.

### DIFF
--- a/datafiles/templates/Html/package-page.html.st
+++ b/datafiles/templates/Html/package-page.html.st
@@ -13,8 +13,8 @@
   <meta property="og:url" content="$baseurl$/package/$package.name$" />
   <meta property="og:site_name" content="Hackage" />
   <meta property="og:type" content="website">
-  <meta property="og:title" content="$package.name$$if(package.optional.hasSynopsis)$: $package.optional.synopsis$$endif$" />
-  <meta property="og:description" content="$package.optional.description$" />
+  <meta property="og:title" content="$package.name$" />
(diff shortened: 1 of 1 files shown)


Kleidukos:
Regarding `og:description`, it is supposed to be "one or two sentences" ([_og:description - A one to two sentence description of your object_](https://ogp.me/)). We do not enforce this in the `description` field, and it feels more appropriate that we fill the metadata with the synopsis.

MorrowM:
Good point.
I also thought of having an `og:image` field with the Hackage logo, but I couldn't get it to display, even after making a png (apparently OpenGraph doesn't support svg).

gbaz:
looks good!

Kleidukos:
@MorrowM I think this would be a nice idea but I'll let @gbaz have the last word on this ;)

Issue #527, opened 2016-08-18
Show automatic and manual flags in separate lists


phadej:
As people shouldn't really care about automatic flags...

gbaz:
How do we know which is which automatically?

phadej:
@gbaz we already know it, e.g. https://hackage.haskell.org/package/semigroups

gbaz:
resolved

PR #963, opened 2021-08-09
Separate out automatic flags into a separate table

MorrowM:
Implements #527

The exact specification wasn't clear, so here is the design I went for: The manual/auto column is removed, and instead the main table consists entirely of the manual flags. Below the main table is a collapsed table containing the automatic flags. Attached is an example.

image

image

Any input on the design is welcome. My only concern is that it may be unclear at first glance that the main table implicitly contains only manual flags.


gbaz:
I'd suggest changing the heading for the main flags table to be "Manual Flags" just to be fully explicit. Other than that, seems good. Thanks!

Issue #748, opened 2018-04-30
move to full-text-search package.


gbaz:
@adamgundry writes (in https://github.com/haskell/hackage-server/issues/615):

It might be worth switching Hackage to use the `full-text-search` package. This was derived from the Hackage search implementation (so it provides a very similar interface and switching should be fairly trivial), but it has a few extra features, notably autocomplete search.

gbaz:
@adamgundry you mentioned at some point an interest in doing this yourself. Gentle ping to see if you're still potentially interested.

Issue #727, opened 2018-03-24
Auto-Complete for Search Box


gbaz:
Good suggestion. The question is what we auto-complete to? I suppose package names, possibly tags...

hvr:
fwiw, we have autocomplete at http://matrix.hackage.haskell.org/ ; that requires to seed the webbrowser with a JSON array of all package names

saevarb:
I think the auto-complete at http://matrix.hackage.haskell.org/ is very nice.

Is it desirable to avoid seeding the browser with all the package names? Looking at the API call that's being made, https://matrix.hackage.haskell.org/api/v1.0.0/package/list/?count=100000, it seems to be limiting it to a 100.000 packages. The returned response is roughly 600 kilobytes of JSON. In the grand scheme of things, it's not a lot(and probably made a lot less by compression), but I think it can still be avoided.

We could simply send an API call upon typing a character, which would then limit the response to packages starting with those characters. In order to avoid hitting the API too much, we could just cache the responses so that if I type `a`, then delete it, and type `a` again, we don't need to hit the API again.

hvr:
fwiw, that's actually the old faked API; the new API (which isn't live yet) just sends a `[Text]` array; and it's *very* cheap server-side to send it (and it's also very CDN friendly, as it uses Etags)

PR #968, opened 2021-08-17
WIP: migrate search engine to full-text-search package (#748)

adamgundry:
See #748. This removes the search engine code in favour of the `full-text-search` package (which was derived from the Hackage codebase originally, so is almost exactly compatible). This is a very quick attempt; I haven't verified if there are code changes on either side that need to be dealt with.

It has a couple of awkward places that still need attention:
 * The `queryExplain` API apparently changed fractionally, which affects `searchPackagesExplain`.
 * I don't know how to define a `MemSize` instance, because `full-text-search` does not expose the internal representation of `SearchEngine`.


gbaz:
Would this let us implement https://github.com/haskell/hackage-server/issues/727 as well?

adamgundry:
> Would this let us implement #727 as well?

It gives one possible strategy, yes.

The `full-text-search` package offers `queryAutosuggest` (https://hackage.haskell.org/package/full-text-search-0.2.1.4/docs/Data-SearchEngine.html#v:queryAutosuggest) which given a query with a partial last term, uses the search index to find potential completions of the term and produce both a ranked list of completions and the corresponding result documents. (I can't remember off the top of my head whether the completions are already stemmed; I think they might be.) Thus it's a bit more flexible than just autocompleting based on package name, but on the other hand the result quality depends on how well-tuned the fields/parameters are in the search index.

We'd then need to expose a query API wrapping `queryAutosuggest`, and have some frontend scripting to issue queries as the user types. I'm not really familiar enough with `hackage-server` to know how easy that would be. And even with debouncing, I guess that could potentially increase load on the server a fair bit?

gbaz:
I think we really should merge this. Sorry for letting it languish. Can you explain what "The queryExplain API apparently changed fractionally, which affects searchPackagesExplain." means in more detail?

(the memsize issue doesn't matter so much, imho)

adamgundry:
> I think we really should merge this. Sorry for letting it languish. Can you explain what "The queryExplain API apparently changed fractionally, which affects searchPackagesExplain." means in more detail?
> 
> (the memsize issue doesn't matter so much, imho)

No need to apologise, because this clearly wasn't quite in a mergeable state when I threw it together, and apologies myself for not getting back to this sooner.

I've looked slightly more closely at the `queryExplain` issue, and I think the issue is that cf0f17ed0a86e52c689f9adbe2fd07940597e3cb changed the search implementation on the `hackage-server` side, but this change isn't (yet) present in `full-text-search`. In particular this needs changes around `query` as well, but its type didn't change, so I hadn't noticed before.

Also, I noticed that `queryExplain` was removed from `hackage-server` in #1021. This seems unfortunate if we want to refine the field weighting in the future? cc @ysangkok

ysangkok:
@adamgundry Since the `queryExplain` code is really a debugging feature, I thought there is no need to keep it around in the codebase. If the search algorithm needs adjusting, these changes can be easily undone, since they were only removing code.

If you think it is better to keep the unreferenced code around, I wouldn't mind too much. I just think it is better to err on the side of removal. To me it is a philosophical choice, one could go either way, but if we choose to keep unused code around, that choice should be explicit so people can take it into consideration.

PR #972, opened 2021-09-02
Remove flag sections when they're empty

MorrowM:
Remove individual flag sections (i.e. auto/manual) when they're empty.

Example: https://hackage.haskell.org/package/cereal

Also, adds an overall "Flags" heading for consistency, in addition to the current "Manual Flags" heading.


gbaz:
thanks!

Issue #948, opened 2021-07-02
usernames should not be easily discoverable


ketzacoatl:
We use our login credentials to access protected functions and so the system knows who we are when performing those functions. Hackage is a database of packages used by the Haskell ecosystem. Package supply chain attacks are a real thing. Protecting packages, is important. As part of that, we need to protect admins (hackage users) who manage those packages. There should be no public list of usernames. If a username must be looked up, a user should have a form where they provide an email address, and if the system has a username associated with that email, an email is sent to the address with the username noted therein. At the very least, we could provide the list of users without their usernames, eg using the "display name". The list of usernames should not be discoverable as they are.

phadej:
I don't understand what concrete changes you are requesting to be done.

Who uploaded packages, and who are the maintainers off a package is important information for the livehood of the ecosystem. (i.e *Uploaded by* and the package maintainers list (which AFAIK there is no direct link)).

If you are worried about https://hackage.haskell.org/users/, I agree, that endpoint is completely useless.

ketzacoatl:
The issue is not about who maintains a given package, but rather the fact that usernames for _all_ users of hackage are so easily discoverable. Dropping the users endpoint would be a great start. The other update is to use the user's display name rather than their username (when referring to the user in packages/etc).

hsek:
Should I keep the users endpoint for admins only, or do I just remove it entirely?

phadej:
> but rather the fact that usernames for all users 

What is a usecase to have an account on Hackage if person is not planning to become a package maintainer? There are probably many such user accounts, but IMO they can be just dropped.

ketzacoatl:
@phadej again, the issue is about how easily discoverable the actual usernames are, for all users, and is relevant for phishing and security and privacy in general.

phadej:
For uploader, it's very simple, index file has everything already:

```
% tar -tzvf /cabal/packages/hackage.haskell.org/01-index.tar.gz|head -n 30
-rw-r--r-- IanLynagh/Hackage 779 2006-09-06 02:37 iconv/0.2/iconv.cabal
-rw-r--r-- IsaacJones/Hackage 4541 2006-09-24 22:15 Crypto/3.0.3/Crypto.cabal
-rw-r--r-- IsaacJones/Hackage  481 2006-09-24 22:15 HDBC/1.0.1/HDBC.cabal
[...]

PR #975, opened 2021-09-16
Remove `/users` GET endpoint

hsek:
I was following the discussion here #948. Although this only contains the dropping of the endpoint.


gbaz:
As per the discussion there, I don't think this is something that brings any particular security benefit. I should add that reviewing the full list of users is sometimes (though not commonly) useful as a hackage admin, so at most i'd imagine guarding this behind a capability check.

hsek:
@gbaz Alright, thanks for taking a look at this still! :)

PR #976, opened 2021-09-16
Update setup instructions

### LINKED ISSUE #974: Readme update
The readme recommends installing via:
```
cabal install -j --enable-tests
```
When I do this I get:
```
cabal: --enable-tests was specified, but tests can't be enabled in a remote
package
```

Would it be equivalent in current `cabal` to do something like:
```
cabal configure --enable-tests
cabal install -j
```

hsek:
Closes #974


gbaz:
thanks!

PR #978, opened 2021-09-21
Ask new uploaders to name the package they want to upload

endgame:
(And to only upload packages they believe will be useful and intend
to maintain.)

Hopefully this will remove one email round-trip from most Hackage uploader requests, and discourage submission of toy packages.

I will email this PR to the Trustees list, so they get a chance to weigh in before merge.

### DIFF
--- a/datafiles/templates/UserSignupReset/SignupConfirmation.email.st
+++ b/datafiles/templates/UserSignupReset/SignupConfirmation.email.st
@@ -6,10 +6,14 @@ this link:
 
   $confirmlink$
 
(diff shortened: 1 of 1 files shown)


endgame:
Maybe we want to add a link to a pre-release checklist somewhere that we can change easily without needing to redeploy hackage? Any suggestions where?

gbaz:
I'll merge this as is, and we can always update later. A pre-release checklist could go either on the haskell wiki, or the trustees wiki, or just in a .md document in the trustees repo?

PR #1002, opened 2022-02-11
update `nix-shell` sources; GitHub Actions for `nix-shell`

peterbecich:
- update `nix-shell` sources
- use `niv` (https://github.com/nmattia/niv)
- GitHub Actions to check `nix-shell`

-----------------------

`nix-shell` builds the project successfully in GitHub Actions and locally:
```
 nix-shell --pure --run "cabal update && cabal build all --enable-tests"    
...
Downloading the latest package list from hackage.haskell.org
Updated package list of hackage.haskell.org to the index-state 2022-02-13T21:11:53Z
To revert to previous state run:
    cabal v2-update 'hackage.haskell.org,2022-02-13T18:44:13Z'
Resolving dependencies...
Up to date
```


gbaz:
Why is `nix/sources.nix` checked in here? Its a big autogenerated file -- I confess I'm not super familiar with this stuff, but its a bit weird to me that such a file would need to be checked in.

peterbecich:
I can delete the Niv https://github.com/nmattia/niv stuff and update the Nix channel manually if you prefer.

A.f.a.i.k. Niv currently requires the `nix/sources.nix` file. It was generated by `niv init`. This is an open issue about getting rid of this file: https://github.com/nmattia/niv/issues/272

In this PR, `shell.nix` imports `nix/sources.nix` and fails without it:
```
nix-shell --pure
error: getting status of '.../nix/sources.nix': No such file or directory
(use '--show-trace' to show detailed location information)
```

I could inline `sources.nix` into `shell.nix` somehow but this would go against the Niv documentation. Probably better to just not use Niv in this case.

-----------

Regardless of Niv, the Nix channel is updated and `brotli` is provided now: https://github.com/haskell/hackage-server/pull/1002/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L24

peterbecich:
I see the "Test nix-shell" GitHub Action takes a long time. I think it does not cache. More discussion about this: https://github.com/actions/cache/issues/749#issuecomment-1059944800
I could try the solution discussed there, or wait for this to possibly be incorporated into https://github.com/cachix/install-nix-action

gbaz:
if you hold off on the niv stuff i'll merge the basic fixes.

shell.nix
```diff
@@ -1,22 +1,26 @@
 let
   nixpkgs = builtins.fetchTarball {
-    # master on 2021-08-02
-    url = "https://github.com/NixOS/nixpkgs/archive/70e001f35cc363eb789ea0a04eff11b86c440ba3.tar.gz";
-    sha256 = "1mrhbcfa8kkx1qnax8xh41grinqiycl56wlws5vvrli8w0pzgl1r";
+    # master on 2022-05-14
+    url = "https://github.com/NixOS/nixpkgs/archive/1d370bd07399fb52cea6badfbffbc90ac9b0f8f0.tar.gz";
```
peterbecich, line 4: https://github.com/NixOS/nixpkgs/commit/1d370bd07399fb52cea6badfbffbc90ac9b0f8f0

PR #1004, opened 2022-02-12
update Hackage Matrix link

ulysses4ever:


ulysses4ever:
Hm, this is sad. I tried it on one package (`async`), and simply changing the URL did seem to show something reasonable (I may be wrong of course).

I think if it shows something reasonable _sometimes_, it's already improvement upon the current state. But I understand (a) it may not, (b) you may disagree. In either case, please, feel free to close this PR.

gbaz:
cc: @davean

ulysses4ever:
MAtrix has long been dead, sadly, so this PR is bust.

PR #1005, opened 2022-02-12
Fix compilation with haddocks

ysangkok:
Can be tested with 'cabal haddock'.

### DIFF
--- a/Distribution/Server/Framework/RequestContentTypes.hs
+++ b/Distribution/Server/Framework/RequestContentTypes.hs
@@ -2,7 +2,6 @@
 -- |
 --
 -----------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


Distribution/Server/Framework/RequestContentTypes.hs
```diff
@@ -2,7 +2,6 @@
 -- |
 --
 -----------------------------------------------------------------------------
--- |
```
andreasabel, line 5: This was the wrong `-- |` to delete.  (The parser for the module header (license etc.) is very fragile.)
I am fixing this.

andreasabel (approved):
Thanks a lot!

PR #1006, opened 2022-02-13
Generate a more complete CI using haskell-ci

andreasabel:
This PR contributes an alternative CI that is built with `haskell-ci` plus some postprocessing via `patch`.

The patching works around the following limitations of `haskell-ci github`:
1. Installing extra ubuntu libraries (in this case `libbrotli-dev`)
2. Activating a haddock step via option (`--haddock`).  A haddock step is currently only added for `library` packages.

The hope with a `haskell-ci` generated CI would be that it is easier upgradable (by upgrading `haskell-ci`, rather than tinkering with the workflow).  It might also be more robust.

But I am not sure if this PR is an overall improvement of the status quo; there are also drawbacks, like extra indirection.

Maybe this PR shouldn't be merged right now, but might be come in handy in the future...

(If merged, then please squash.)


.github/workflows/haskell-ci.patch
```diff
@@ -0,0 +1,20 @@
+--- haskell-ci.yml.orig	2022-02-13 11:28:52.000000000 +0100
++++ haskell-ci.yml	2022-02-13 11:37:48.000000000 +0100
+@@ -43,6 +43,7 @@
+       - name: apt
+         run: |
+           apt-get update
++          apt-get install -y --no-install-recommends libbrotli-dev
```
phadej, line 7:
```
% haskell-ci --help|grep apt
                    --ghcup-version VERSION | --apt PKG | 
  --apt PKG                Additional apt packages to install
```

andreasabel:
Superseded by
- #1061 

thanks to 
- https://github.com/haskell-CI/haskell-ci/pull/582
- https://github.com/haskell-CI/haskell-ci/pull/585

Issue #1001, opened 2022-02-10
Email address detection in markdown faulty


andreasabel:
See e.g. [hackage.haskell.org/package/BNFC-2.9.4/candidate](https://hackage.haskell.org/package/BNFC-2.9.4/candidate):
The email bnfc-dev@googlegroups.com (under Support) is rendered wrongly as bnfc[-dev@googlegroups.com](mailto:-dev@googlegroups.com).

gbaz:
I believe this issue is with `commonmark` that hackage uses for markdown.

andreasabel:
Installing the latest version, with `commonmark -x autolinks README.md` I get the correct rendering:
> ... our mailing list <a href="mailto:bnfc-dev@googlegroups.com">bnfc-dev@googlegroups.com</a>.

Without the `autolinks` extension, I get no rendering of the mail address:
> ... our mailing list bnfc-dev@googlegroups.com.

andreasabel:
@gbaz Could it be the ordering of syntax specifications, see comment https://github.com/haskell/hackage-server/pull/904/files#r804234727 ?

Issue #1000, opened 2022-02-10
Plain http(s)-links in markdown not rendered


andreasabel:
See e.g. https://hackage.haskell.org/package/BNFC-2.9.4/candidate: 
Links like https://bnfc.readthedocs.org/en/latest/ (under heading Documentation) are not rendered as links.  (Note that github does render such links.)

andreasabel:
@gbaz: To see the change, I suppose the server needs to be relaunched.  When does this happen (and who does it)?

PR #1013, opened 2022-02-14
Doc tests using `doctests-driver-gen`

andreasabel:
This contributes:
- a few doctests, testing 
  * #1000 
  * #1001
- a new test-suite `DocTests` running these tests

This required me to:
1. depend on `doctest` and `doctest-driver-gen`
2. move the library code into `src`:
  I cannot pass `.` (where currently the library sits) to `doctest`, as this would pick up also the `exes`, and they all define module `Main`, which of course gives ambiguous module names and is not liked by `ghci`.
3. add `write-ghc-environment-files: always` to `cabal.project`.

Step 2. is maybe a big change, but it is in a single commit, so rebasing PRs should work (I rebased a bit back and forth over this commit locally, and it went fine).  Putting the library in a subdir is the standard practice, so in the long run we have to do it at some point.

How to run the doctests:
```
cabal build --enable-tests DocTests
cabal test --enable-tests DocTests --test-show-details=direct
```
The first step writes the `.ghc.environment` file which is needed by `cabal test`.
One could skip writing the environment file, but then it would need to be prefixed by `cabal exec`:
```
cabal exec -- cabal test --enable-tests DocTests --test-show-details=direct
```

As usual, running `doctest` is very slow (30 sec on my mac).  I also tried `doctest-parallel` in #1014 (https://github.com/haskell/hackage-server/blob/parallel-doctest/tests/DocTestMain.hs), but I didn't get it to work, and also it did not appear too much faster (17 sec getting to the point where it failed).

### DIFF
--- a/hackage-server.cabal
+++ b/hackage-server.cabal
@@ -561,3 +561,15 @@ test-suite HashTests
       -- component-specific dependencies
     , tasty       ^>= 1.4
     , tasty-hunit ^>= 0.10
+
+test-suite DocTests
+  import:         test-defaults
+
+  type:           exitcode-stdio-1.0
+  main-is:        DocTestMain.hs
+
+  build-depends:  doctest ^>= 0.20
+  build-tool-depends:
+    doctest-driver-gen:doctest-driver-gen
+
+  ghc-options:    -with-rtsopts=-N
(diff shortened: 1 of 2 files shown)


gbaz:
I think moving the library into a subdir is worth it on its own. This historical quirk of the hackage codebase (reflecting its age) has always been a bit irritating.

andreasabel:
> I think moving the library into a subdir is worth it on its own. This historical quirk of the hackage codebase (reflecting its age) has always been a bit irritating.

Ok, I am pushing this part then.

andreasabel:
Closing in favor of #1014.

PR #1014, opened 2022-02-14
Doc tests using `doctest-parallel`

andreasabel:
This is an alternative to #1013, ~~but it does not work (yet?).~~

~~Potential~~ advantages:
- `doctest-parallel` is faster
- ~~no need to move library to `src/`~~

UPDATE: moving library sources into `src/` made this variant work.  It is faster than plain `doctest` (#1013), so I think I will settle for this alternative.

### DIFF
--- a/hackage-server.cabal
+++ b/hackage-server.cabal
@@ -564,3 +564,12 @@ test-suite HashTests
     , tasty       ^>= 1.4
     , tasty-hunit ^>= 0.10
 
+test-suite DocTests
+  import:         test-defaults
+
+  type:           exitcode-stdio-1.0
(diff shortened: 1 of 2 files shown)


martijnbastiaan:
@andreasabel I've release `doctest-parallel` 0.2.2: https://hackage.haskell.org/package/doctest-parallel-0.2.2. If all is well you can remove the workaround.

andreasabel:
@martijnbastiaan: Works like a charm!  Thanks!

andreasabel:
CI is failing, but it is not my fault: https://github.com/haskell/hackage-server/pull/996#issuecomment-1053653226
The `DocTests` pass.

PR #1022, opened 2022-02-22
Use flexbox instead of float

### LINKED ISSUE #928: Visual issues on narrow screens

ysangkok:
Fixes #928.

Instead of using `float: right`, which is designed for content that can be divided in the middle (e.g. floating an image as on Wikipedia), this PR uses flexbox to fit the columns next to each other if possible, but otherwise lets the properties pane go down below if either of description/badges/modules/flags/downloads/maintainer-corner are too wide.

This makes it such that very long module paths will leave the properties pane with no space, and it will be sent down.

I note that the issue #928 is named after narrow screens, but in fact really narrow screens won't exhibit the problem, since they won't have any "float" applied. The issue is with the float, which is what this PR replaces. The floating removed in this PR was only applied on screens with a certain *minimum* screen width.

To test this PR, I'd recommend opening up any package page and using the inspector to make a module name longer. Before this PR, you'd get the weird behavior screenshotted in the linked issue. But with this PR, that doesn't happen.

Note that the diff is much smaller if you ignore white-space! There is a Github setting for that: Click on the cog right below the "Files changed" tab, while being on that tab.

The reason for putting all of these sections in the new `left-pane` is because the README marks the end of tools and Hackage metadata, and it therefore is a good dividing line. There are many small packages with a properties page that is much taller than all of the sections in the `left-pane` put together, as shown in this screenshot (showing behaviour with this PR applied):
image

If the sections were not in the left pane, the white space shown in the screenshot would be even taller.

Tested on Firefox 91 and Chrome 98.

Tagging @noughtmare

### DIFF
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -312,6 +312,7 @@ table.properties td, table.properties th {
 
 @media only screen and (min-width: 1920px) {
   #content {
+    /* keep this in sync with #left-pane children */
     width: 60vw;
   }
 }
@@ -375,18 +376,28 @@ table.properties td, table.properties th {
       margin-bottom: 2em;
   }
 
-  #left-pane {
-      margin-right: 2em;
-      /* The margin is subtracted. Without this line, a narrow left-pane would
-       * leave a narrow properties pane with space on the right. So we make the
-       * left-pane take the space that the properties pane is not allotted.
-       * That leaves no space on the right. */
-      width: calc(60% - 2em);
+  #left-pane > * {
+      /*   (100% - (width of properties)) * (width of #content) - gap
+       * = (100% - 40%                  ) * 60vw                - 2em
+       * = 60%                            * 60vw                - 2em
+       * = 36vw - 2em
+       */
+      width: calc(36vw - 2em);
+  }
+
+  #left-pane > #modules {
+      /* This overrides the width in the previous block. We want a long module
+       * name to make the left-pane wide, so that it will push down
+       * #properties.
+       */
+      width: fit-content;
   }
 
   #flex-container {
       display: flex;
       flex-wrap: wrap;
+      justify-content: space-between;
+      gap: 2em; /* keep this in sync with calculation above */
   }
 }
 
@@ -843,10 +854,6 @@ ul.links li form button {
 
 #index td + td {
   padding-left: 1em;
-}
-
-#module-list {
-  overflow-x: auto;
 }
 
 #module-list ul {
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -381,7 +381,7 @@ table.properties td, table.properties th {
        * leave a narrow properties pane with space on the right. So we make the
        * left-pane take the space that the properties pane is not allotted.
        * That leaves no space on the right. */
-      max-width: calc(60% - 2em);
+      width: calc(60% - 2em);
   }
 
   #flex-container {
@@ -845,19 +845,13 @@ ul.links li form button {
   padding-left: 1em;
 }
 
+#module-list {
+  overflow-x: auto;
+}
+
 #module-list ul {
   list-style: none;
   margin: 0 0 0 20px;
-
-     /* Warning: Needed for oldIE support, but words are broken up letter-by-letter */
-    -ms-word-break: break-all;
-    word-break: break-all;
-    /* Non standard for webkit */
-    word-break: break-word;
-    -webkit-hyphens: auto;
-    -moz-hyphens: auto;
-    -ms-hyphens: auto;
-    hyphens: auto;
 }
 
 /* Disable identation for the top level modules */
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -381,7 +381,7 @@ table.properties td, table.properties th {
        * leave a narrow properties pane with space on the right. So we make the
        * left-pane take the space that the properties pane is not allotted.
        * That leaves no space on the right. */
-      min-width: calc(60% - 2em);
+      max-width: calc(60% - 2em);
   }
 
   #flex-container {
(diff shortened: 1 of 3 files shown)


noughtmare:
I notice that this moves the right panel to the bottom a bit too soon. It seems like it doesn't want to wrap the package description, but I think that would be fine. It is just the module names that shouldn't be wrapped.

ysangkok:
@noughtmare I added a commit that I think resolves the problem. I was able to reproduce your issue on 4Blocks (which means my screenshot above wasn't current when I submitted the PR, sorry about that).

noughtmare:
Now it also starts wrapping the module names again.

ysangkok:
Ok, I have resorted to `width` instead of `min-width` or `max-width`. The issue with only a `min-width` is that packages with a long description will make the properties pane get pushed down. Only `max-width` has the problem that a narrow `left-pane` will leave too much space on the right. So now I use `width`.

To resolve the problem with broken words, I have added `overflow-x: auto` only for the modules section, which makes a scroll bar appear on these edge cases that have very long module names. If the properties pane is pushed down, the left-pane gets more space, so the scroll bar probably won't be necessary.

image

noughtmare:
Ideally, I would like to see the behaviour that the right pane would get pushed down before the scroll bar appears.

ysangkok:
I agree that that would be nice, but I don't know how to achieve that. The flexbox must have only two immediate elements, otherwise I wouldn't know how to make the left-right becomes up-down behavior. But description and module list are contained in the same immediate flexbox child, the flexbox can't discern whether its request for space is illegitimate (description is more than just one line) or legitimate (module name is long).

But I do think this is an improvement over the current state of affairs. Savvy users will know that they can scroll on the x axis using the shift key and their mouse wheel. So the scroll bar can actually be used even though it cannot be seen.

noughtmare:
The left pane scrollbox also sometimes captures my vertical scrolling that happens when hovering over that part of the page.

Also, just switching to the separate layout way earlier (maybe around a total screen width of 1600 px or at least 1200 px) would already be a much bigger improvement than this change for me personally. Or now with your flexbox I would say the left pane should have a minimum width of around 700 pixels.

ysangkok:
Ok, I have found a way to have only the modules section force the `left-pane` to grow in size. I have tested for the scenarios that caused me to do the previous commits:

* ✅ On a page with no modules section, but a long description, the properties pane is on the right 
* ✅ On a "normal" page with a short module names, the properties pane is at the right, and not further towards the middle/left (achieved using `space-between` now). That is, there is no empty space on the right margin.
* ✅ On a module page with super long names, the properties section is pushed down (no scroll bar any more)

I believe these were the requirements you desire. Can you please confirm @noughtmare ?

gbaz:
I will note that something like this really needs to be tested on Chrome,
Firefox, and also Safari.

ysangkok:
I realized that since the width of `#content` changes, the `#left-pane > *` widths should also change. Pushed a commit to address that.

@gbaz I have already tested on Chrome and Firefox (as noted in the initial PR description). But I managed to get hold of my girlfriend's iPad Air model MD787E/A with operating system version 12.5.5. Looks like it is from around the year 2013 (good, since maybe it verifies that I don't use any features that are too new). I have done some tests in landscape mode and in portrait mode, with a package with a super long module name and one with the 4Blocks package which doesn't expose any modules (and has a short description, so it is narrow). It seems to work as expected. I think it is the default configuration, but in the "Experimental Features" menu, all CSS options have been disabled. I'd consider this a good sign, since that means it works as expected even with Experimental CSS features off. The disabled options are "CSS Custom Properties and Values API", "CSSOM View Scrolling API", "CSS Painting API", "CSS Typed OM", "CSS Spring Animations".

One thing to note during testing is how the distance between the black line and the properties pane is always around 2em, which it wasn't before the last commit, this has been rectified.

The last screenshot shows how even when the properties pane goes down below, content in the `left-pane` that is narrow will have a black line that doesn't extend all the way to the right. I don't think that is a real problem though, it looks better if the line has a length comparable to the elements around it.

newproj-portrait
newproj-landscape
usual-portrait
usual-landscape

andreasabel:
While you are discussing this, please have a look at
- https://github.com/quchen/prettyprinter/issues/192

where I was fooled by the Hackage layout of the module tree.  Would be great if such misleading display would be fixed by this PR.

ysangkok:
@andreasabel I was able to reproduce your problem on a full-screen Firefox on Ubuntu 20.04 with `xdpyinfo` reporting

    dimensions:    3200x1800 pixels (846x476 millimeters)
    resolution:    96x96 dots per inch

image

Indeed with this PR the long module name has pushed down the properties pane:

image

So I'd say this PR also solves #192.

noughtmare:
@andreasabel @gbaz can this be merged?

andreasabel:
> @andreasabel @gbaz can this be merged?

I am afraid I lack competences to do so (although I have commit rights).  I have not insight into this part of the code, and CI (unsuprisingly) does not test UI features.
Also, I don't know how "owns" `hackage-server` and has the final responsibility (and makes the decisions that are not obvious).  Is it just you now, @gbaz?  (That's what I conclude from the activity on this repo.)

noughtmare:
Thanks, I also didn't know who the best person to ask for this is. I guess @NunoAlexandre has written or at least edited most of this stylesheet, so maybe he can review these changes? And I think `hackage-server` is owned by the [infrastructure admins](https://github.com/haskell-infra/haskell-admins#the-team-and-how-to-contact-them) (which includes gbaz).

gbaz:
I ended up being the central maintainer, along with davean. I do intend to review these changes in a few browsers and smoke test them and then merge before the next hackage redeploy. Unfortunately, I've been busy with some other work so haven't gotten to it yet.

PR #1024, opened 2022-02-23
assume a default content type of text/html when negotiating content

gbaz:
should help with the test failure on https://github.com/haskell/hackage-server/pull/996

I can't see any potential downside here. Modern http clients will almost certainly send an accepts header. The current behavior is nonuniform because the "default" varies by whatever content type is listed first. And I think it is most in keeping with the spirit of the old web to assume people want html unless they specify otherwise.

That said I'll let it sit for a few days before merging and see if there is comment.


gbaz:
This change is already included in the above mentioned pr so closing.

PR #1008, opened 2022-02-13
support for `prefers-color-scheme`

peterbecich:
Closes https://github.com/haskell/hackage-server/issues/1007
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

Some of the foreground and background colors are simply swapped. ~`href` colors should probably be brightened.~ Opinions appreciated. Preview: https://github.com/haskell/hackage-server/pull/1008#issuecomment-1138132085

PR #1025, opened 2022-02-27
update DataTables

peterbecich:
This PR updates DataTables from 1.10.12 to 1.11.4. Is it okay to update this?

I've found it necessary to update DataTables to complete https://github.com/haskell/hackage-server/pull/1008
Specifically, I think the older version of DataTables currently used does not support setting the text color of the "SearchPanes" extension (Show 10 entries ... Filter).

The source is https://datatables.net/download/ ; the styling framework is "DataTables default styling," the package is "DataTables," and one enabled extension is "SearchPanes." No other options are selected.

The PR deletes the current version of DataTables from version control, and uses the DataTables CDN instead. (For jQuery https://github.com/haskell/hackage-server/pull/1026)

I think the result is identical:
Screenshot 2022-02-26 at 21-50-53 All packages by name Hackage


gbaz:
Datatables will be subsumed in 90% of cases by the new browse functionality which uses something else. As such I'm fine switching to a cdn for this since its not going to be core functionality.

PR #1026, opened 2022-02-27
use jQuery CDN

peterbecich:
The version of jQuery is unchanged, 1.12.4. The CDN (_content delivery network_) was found here: https://developers.google.com/speed/libraries#jquery


ysangkok:
What is the advantage of using an additional CDN? Hackage is already served from a CDN. With this change, users that use script blocking extensions like uMatrix would need to allow an additional domain.

peterbecich:
I hadn't thought of that, but `fonts.googleapis.com` and `jsdelivr.net` are already used:
https://github.com/haskell/hackage-server/blob/d69f7b69852d2e81313753d730f17eeb06254e97/datafiles/templates/Html/candidate-page.html.st#L8
https://github.com/haskell/hackage-server/blob/d69f7b69852d2e81313753d730f17eeb06254e97/datafiles/templates/hackageCssTheme.st#L2
https://github.com/haskell/hackage-server/blob/d69f7b69852d2e81313753d730f17eeb06254e97/src/Distribution/Server/Pages/Template.hs#L67
https://github.com/haskell/hackage-server/blob/d69f7b69852d2e81313753d730f17eeb06254e97/datafiles/templates/Html/package-page.html.st#L23


The only two reasons I can think of to merge this PR are:
- reuse jQuery 1.12.4 pre-cached by visiting other websites
- delete lines of code from this repository

nickburlett:
If you do this, please make it configurable... Having it use a CDN will prevent a hackage mirror from working in an air-gapped system.

gbaz:
I think if the fonts or mathjax fails that just screws up rendering. We might actually use the jquery functionality for something important to user experience.

peterbecich:
Closing; I could not make this work with a fallback option.

PR #1031, opened 2022-03-14
GHC 9.0 and Cabal 3.4 (fix #1016)

### LINKED ISSUE #1016: GHC 9.0 and Cabal 3.4
This issue tracks progress of building with GHC 9.0.  Add blockers to the list:
- [x] `hackage-security` (https://github.com/haskell/hackage-security/pull/265, https://github.com/haskell/hackage-security/pull/266)
- [x] `acid-state` (https://github.com/acid-state/acid-state/issues/137)

Building can now be started with:
```
$ cabal build -w ghc-9.0.2 --allow-newer=Cabal --constraint='Cabal==3.4.*'
```
TODO:
- [ ] migrate to `Cabal-3.4`

andreasabel:
Migrating to GHC 9.0 needs (one commit):

- reordering declarations so that Template Haskell splices have everything in scope when they execute
- space after `!` and `$`
- putting constraints (`C =>`) before domains (`A ->`) in type signatures

Migration to Cabal 3.4 has these components (each one commit):
- `VersionRangeF`: some cases removed
- `Distribution.Server.Util.CabalRevisions`: copied from `hackage-cli`
- `mainLibSet`
- `Flag` -> `PackageFlag`
- need to import `Distribution.Types.ModuleReexport` explicitly
- `RepoType` got intermediate layer `KnownRepoType`:
  * [x] Do we add new code for new repo types?
- `FieldGrammar`
  * [ ] Do we need to add fields?
- `Version` to `CabalSpecVersion`: largest change, needs review.
  * [x] Is the code in `Distribution.Server.Packages.Unpack.specVersionChecks` still needed/meaninful?  It does not mention `cabal-version: 3.0` or higher.

- [x] This warning should be investigated:
```
src/Distribution/Server/Features/PackageCandidates.hs:209:42: warning: [-Wmissing-fields]
    • Fields of ‘CoreResource’ not initialised: coreCabalFileRev
    • In the expression:
        CoreResource
          {corePackagesPage = (resourceAt "/packages/candidates/.:format")
          ...
```

### DIFF
--- a/src/Distribution/Server/Features/BuildReports/BuildReport.hs
+++ b/src/Distribution/Server/Features/BuildReports/BuildReport.hs
@@ -1,6 +1,12 @@
-{-# LANGUAGE DeriveDataTypeable, GeneralizedNewtypeDeriving, RecordWildCards,
-             TemplateHaskell, TypeFamilies, FlexibleInstances, MultiParamTypeClasses,
-             OverloadedStrings #-}
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeFamilies #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module      :  Distribution.Client.Reporting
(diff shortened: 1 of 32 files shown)


gbaz:
because i'm doing some devel on an m4, i find myself compelled to get it working locally with 9.2.2 -- ok if I push my changes for that to this branch?

Also, any clue as to the failing checks??

gbaz:
Looks like the failures may have been spurious timeouts... hrm.

gbaz:
We should test but I'm pretty sure specversioncheck is indeed used and should probably be bumped somewhat.

andreasabel:
> because i'm doing some devel on an m1, i find myself compelled to get it working locally with 9.2.2 -- ok if I push my changes for that to this branch?

Yes, sure.  
But don't you have to include the upgrade to Cabal 3.6 then?  If this is the case, maybe this would be better in a second step.

> Also, any clue as to the failing checks??

I think it is #1032.

gbaz:
Pushing a patch with some changes. Not doing the 9.2.1 thing because I'm back on a different box. Added a dummy thing to clean up the warning. Deleted some stray repotype code and commented what types are still missing. More generally in fact I took the opportunity to clean out some other dead code from the Package.hs file.

Cleaned up and bumped the specversion check since the strict enumeration lets us simplify the logic.

I'm not sure what the question "Do we need to add fields?" for FieldGrammar refers to.

gbaz (approved):

src/Distribution/Server/Features/BuildReports/BuildReport.hs
```diff
@@ -308,7 +317,21 @@ show = showFields (const []) . prettyFieldGrammar CabalSpecV2_4 fieldDescrs
...
+       , c (Identity OS)
+       , c (Identity Outcome)
+       , c (Identity PackageIdentifier)
+       , c (List VCat (Identity PackageIdentifier) PackageIdentifier)
+       , c Time
+       )
+    => g BuildReport BuildReport
 fieldDescrs =
   BuildReport
     <$> uniqueField       "package"            packageL
```
andreasabel, line 337: Do we need more fields here?

andreasabel:
> I'm not sure what the question "Do we need to add fields?" for FieldGrammar refers to.

If I remember correctly, I wondered whether there would be new fields required here:
https://github.com/haskell/hackage-server/pull/1031/files#r834770535
But probably not, as this is not a cabal file thing, but a hackage server thing (thing = data structure).

gbaz:
Indeed, that's just reusing the FieldGrammar stuff to whip up a parser for build reports, which are just about hackage server.

PR #1042, opened 2022-03-29
Add support for embedding of analytics pixels on package pages

alexbiehl:
## Summary

This PR adds support for package maintainers to attach analytics pixels to one or more of their packages in order to better understand the web traffic to their package pages on Hackage. Specifying an analytics pixel URL will result in Hackage automatically embedding the provided URL in an image tag on the main package page. 

## Context

This PR comes from a previous discussion that @aviaviavi, @gbaz, and the hackage-admin email group. The original suggestion for the direction in this PR came @davean, in discussion with the Haskell Foundation regarding approaches for introducing more analytical tools for maintainers within Hackage.

## What not in this PR

Support for analytics pixels to be embedded in generated Haddock documentation is a natural next step and can be included in a future PR.

### DIFF
--- a/datafiles/templates/Html/analytics-pixels-page.html.st
+++ b/datafiles/templates/Html/analytics-pixels-page.html.st
@@ -1,0 +1,45 @@
+<!DOCTYPE html>
+<html>
+<head>
+$hackageCssTheme()$
+<title>Analytics pixels for $pkgname$ | Hackage</title>
+</head>
+
+<body>
+$hackagePageHeader()$
+
+<div id="content">
+
+<h2>Adding a analytics pixel to <a href="/package/$pkgname$">$pkgname$</a></h2>
+
+<p> 
+  Configure an analytics pixel to be automatically loaded on your package’s page on Hackage. 
+  You’ll need an image URL from any external analytics provider, which is provided 
+  for free and can surface information about web traffic to your package including geographic 
+  distribution, version distribution, and companies.
+</p>
+
+<form method="POST" class="box" action="/package/$pkgname$/analytics-pixels">
+    <label for="analytics-pixel">Analytics Image URL</label>
+    <input name="analytics-pixel" type="text" />
+    <input type="submit" />
+</form>
+
+<h2>Existing analytics pixels for <a href="/package/$pkgname$">$pkgname$</a></h2>
+
+<ul>
+  $analyticsPixels:{analyticsPixel|
+    <li>
+        <form method="POST" action="/package/$pkgname$/analytics-pixels">
+            <label for="analytics-pixel">$analyticsPixel$</label>
+            <input type="hidden" name="analytics-pixel" value="$analyticsPixel$"/>
+            <input type="hidden" name="_method" value="DELETE" />
+            <input type="submit" value="Delete" />
+        </form>
+    </li>
+  }; separator=""$
+</ul>
+
+</div>
+</body>
+</html>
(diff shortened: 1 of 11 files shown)


src/Distribution/Server/Features/AnalyticsPixels/State.hs
```diff
@@ -0,0 +1,109 @@
...
+    deriving (Show, Eq, Ord, NFData, Typeable, MemSize)
+
+newtype AnalyticsPixelsState = AnalyticsPixelsState
+    {
+        analyticsPixels :: Map PackageName (Set AnalyticsPixel)
+    }
+  deriving (Show, Eq, NFData, Typeable, MemSize)
+
+-- SafeCopy instances
+$(deriveSafeCopy 0 'base ''AnalyticsPixel)
```
andreasabel, line 48: GHC 9.x's TemplateHaskell seems to process the source top-to-bottom, so the TH invocations need to be put in the right locations.  See the reorderings I performed in 9d4a4900584a87f85f3ca72a8fc907dcfbc1e21b when migrating `hackage-server` to GHC 9.0.

andreasabel (changes requested):
Building with GHC 9.x needs code rearrangements (due to TH).

src/Distribution/Server/Features/AnalyticsPixels/State.hs
alexbiehl: Thanks for the pointer! Will take care of it

alexbiehl: Done!

datafiles/templates/Html/analytics-pixels-page.html.st
```diff
@@ -0,0 +1,45 @@
...
+
+<div id="content">
+
+<h2>Adding a analytics pixel to <a href="/package/$pkgname$">$pkgname$</a></h2>
+
+<p> 
+  Configure an analytics pixel to be automatically loaded on your package’s page on Hackage. 
+  You’ll need an image URL from any external analytics provider, e.g. <a href="about.scarf.sh">Scarf</a>, which is provided 
+  for free and can surface information about web traffic to your package including geographic 
+  distribution, version distribution, and companies.
```
gbaz, line 19: This text, which explicitly directs users to a specific service and provides promotional copy for it, is not going to work. We can merge without changes here, and I'll just modify it in a subsequent commit, if you prefer. But just wanted to be upfront about that. Ditto for the same text in the other template.

src/Distribution/Server/Features/Upload/State.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE DeriveDataTypeable, TypeFamilies, TemplateHaskell #-}
+{-# LANGUAGE DeriveDataTypeable, RecordWildCards, TypeFamilies, TemplateHaskell #-}
```
gbaz, line 1: I think adding this reverse index is a rather heavyweight and unnecessary change. Fetching all packages for a user is already sufficiently efficient. Cf https://hackage.haskell.org/user/BryanOSullivan -- fast!

alexbiehl: OK! I will change that then.

alexbiehl: I updated the PR using the GroupIndex mechanics and reverted all the changes to the Upload feature.

datafiles/templates/Html/analytics-pixels-page.html.st
alexbiehl: I removed the references!

alexbiehl:
Any chance for another round of reviews @gbaz?

gbaz:
It looks good. I  plan to give it a final review and merge before the next redeploy. Just been a bit slammed by other responsibilities!

gbaz:
I merged in preperation for a redeploy, but noticed in testing that the list of pixels already created did not show up properly in the package page. (It does show up in the page that lists all pixels for all packages for a user). So I'm going to comment out links to this functionality for now in the release. A patch fixing that up would be welcome.

PR #1047, opened 2022-04-03
Add form to update full name and email of user

### LINKED ISSUE #167: Account update page
Currently, the only way to update a user's name or, more importantly, email address, seems to be a PUT to /user/<name>/name-contact with the JSON. It would be useful to have an actual html page where users can update their email address (probably with another confirmation email) and user name.

ysangkok:
Fixes #167. Note how the existing server endpoint is used, so this is only adding a front-end for existing functionality.

### DIFF
--- a/datafiles/templates/UserDetails/user-details-form.html.st
+++ b/datafiles/templates/UserDetails/user-details-form.html.st
@@ -1,0 +1,34 @@
+<!DOCTYPE html>
+<html>
+<head>
+$hackageCssTheme()$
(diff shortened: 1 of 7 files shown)


ysangkok:
@gbaz I have added pushed additional improvements:

* The email can now only contain one at-sign, even during signup
* The user details form now uses the same validation that the signup form uses
* The user details form no longer allows you to submit a blank email, which would be rejected by the now-common server side email validation anyway

Issue #1051, opened 2022-04-06
Hackage's search and package pages are no longer accessible


Merivuokko:
A recent update broke Hackage's accessibility. Search and package pages no longer work with browsers which do not have support for JavaScript execution. Such browsers include Lynx and Elinks, which are commonly used by blind users.

Possible workarounds include: 1) provide the original content of these pages within a <nosciprt/> section, and 2) provide separate accessible pages for non-JS browsers.

On Unix-derivatives blind users are not able to switch over to JS-supporting browsers, as the graphical screen reading software on these platforms is inferior.

To date hackage has been an exceptionally accessible site, and I wish to see this support continued.

gbaz:
cc: @ysangkok

Merivuokko:
I tested the new pages with Lynx (I previously used Elinks) and I found out, that the package pages work with Lynx. So with package pages the issue is something else than JS – probably a redirection issue.

If I direct Elinks to https://hackage.haskell.org/package/containers, I et the following kind of JSON as a response:

```json
{"0.1.0.0":"normal","0.1.0.1":"normal","0.2.0.0":"normal","0.2.0.1":"normal","0.3.0.0":"normal", ...}
```

If I load the same URL with Lynx, I get a warning:

```
Alert!: HREF in BASE tag is not an absolute URL.
```

But then the page loads fine.

gbaz:
I think the former issue (elinks) is because there is now a json content type and hackage server will default to it _in this specific instance_ if a browser does not specify a desired content type. I find it weird that elinks doesn't send a content-type tag, but perhaps this should encourage us to enforce that the "default content type" is always text/html if it exists.

Not sure about the base tag issue -- prs welcome.

Vis a vis putting the old code on alternative endpoints -- I suppose this is ok. I would much prefer we had a path that enabled new search but computing the results on the server side rather than the browser, but I understand if that would be rather too much work...

ysangkok:
The HTML versions of the package pages did not have any JavaScript added, in fact a bit of JavaScript was removed as part of #1022. But you are correct that the browse and search pages now require JavaScript. This could be restored as previously mentioned.

Here is how to reproduce the default content-type issue pointed out by @gbaz:

```
$ curl -s -D - "https://hackage.haskell.org/package/containers" | grep -i Content-type
Content-Type: application/json
$ curl -s -D - -H "Accept: text/html" "https://hackage.haskell.org/package/containers" | grep -i Content-type
[...]

PR #1053, opened 2022-04-09
Default to HTML handler, not the first

ysangkok:
Before this commit, the first handler would be chosen as the default, in the absense of an Accept header from the client.

This means that Elinks gets the JSON version of the package page, because it
doesn't send an Accept header, this was reported by @Merivuokko in [a comment](https://github.com/haskell/hackage-server/issues/1051#issuecomment-1090669842).

After this commit, the package page is again defaulting to HTML:

    % curl -s -D - "http://localhost:8080/package/2captcha" | grep -i Content-type
    Content-Type: text/html; charset=utf-8

### DIFF
--- a/src/Distribution/Server/Framework/Resource.hs
+++ b/src/Distribution/Server/Framework/Resource.hs
@@ -464,8 +464,14 @@ serveResource errRes (Resource _ rget rput rpost rdelete rformat rend _) = \dpat
                 Just answer -> handleErrors (Just format) $ answer dpath
                 Nothing -> mzero -- return 404 if the specific format is not found
                   -- return default response when format is empty or non-existent
-            _ -> do (format,answer) <- negotiateContent (head res) res
-                    handleErrors (Just format) $ answer dpath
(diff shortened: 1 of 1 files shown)


gbaz:
Great sleuthing, thanks!

I confess to being a bit confused why the prior fix was not sufficient?

ysangkok:
@gbaz To sum up what I tried to say on IRC: I believe that [the previous defaulting](https://github.com/haskell/hackage-server/pull/996/files#diff-e472f8254e83d4ed34adcac7f603ab69eda10c76706a2e3b36fa27cbde675d9eR524) was not correct, since it doesn't make sense to me to have both the `def` argument and the defaulting to `text/html` inside `negotiateContent`. So if I can get some acknowledgement that this indeed a mistake, I can try and make a new PR that undos the change from 996, now that this PR has introduced (what I believe) to be a more reliable defaulting. A defaulting which should also work with the previous code. I can attempt to write a unit test to verify this.

Issue #1029, opened 2022-03-07
Recent uploads barely covers 2 days


andreasabel:
Owing to the popularity of Haskell and Hackage, the page https://hackage.haskell.org/packages/recent , listing the 25 most recent uploads, barely covers two days of uploads.

Possible improvements:
1. (Easy but not great:) Show uploads of last 2 weeks.
2. (More work:) Make page configurable with
    a. # per page option
    b. jump to next page button

gbaz:
In the rss description it says it is 20 most recent or the last 48 hours, whichever is greater.

PR #1055, opened 2022-04-12
Solves #1029 - Adds paging to recent packages and recent revisions

LeviButcher:
Implements #1029 

Everything Included:
- [x] Paging support via queryParams (page, pageSize) on recent packages and recent revision (html, rss)
- [x] Extract HTML and RSS responses from RecentPackagesFeature to HTMLFeature
- [x] Uses same paging functionality as the browse/search page but without JS (I also updated it to remove Next/Prev if there isn't one)

Possible Issues:
- I'm not for sure if I handled the cache correctly in the RecentPackagesFeature
- I have it paging over all the data, I could change it to only page over the last 48 hours worth if desired
- Some of the support functions I made are not necessarily safe (Usage of head) but I accounted for errors as much as possible

Let me know of any improvements I can make. :+1:


datafiles/static/browse.js
```diff
@@ -303,7 +303,8 @@ const createPaginator = () => {
     pag.appendChild(createPageLink(maxPage));
   }
   const isNowOnLastPage = state.page === maxPage;
-  pag.appendChild(createPrevNext(state.page + 1, isNowOnLastPage, "Next"));
+  
+  if(!isNowOnLastPage) pag.appendChild(createPrevNext(state.page + 1, "Next"));
```
ysangkok, line 307: I think there is value in showing a disabled button, since it will result in a more consistent layout when viewing different pages. What is the motivation for changing this?

datafiles/static/hackage.css
```diff
@@ -1085,11 +1084,12 @@ a.deprecated[href]:visited {
   background: linear-gradient(to bottom, #585858 0%, #111 100%);
 }
 
-.paginator span {
-  padding: 0 1em;
+.paginator a[href][disabled] {
```
ysangkok, line 1087: If disabled buttons are no longer shown, this style sheet is unnecessary.

src/Distribution/Server/Pages/Recent.hs
```diff
@@ -120,32 +196,30 @@ recentRevisionsURL :: URL
 recentRevisionsURL = "/packages/recent/revisions.html"
 
 
-recentFeed :: Users -> URI -> UTCTime -> [PkgInfo] -> RSS
-recentFeed users hostURI now pkgs = RSS
+recentFeed :: PaginatedConfiguration -> Users -> URI -> UTCTime -> [PkgInfo] -> RSS
```
ysangkok, line 199: I don't know if this is too much work, but there is [a standard for paginated feeds](https://www.rfc-editor.org/rfc/rfc5005#section-3). Seems to me like there is no way for the feed reader to show the next page using the current iteration of this PR, is that correct?

datafiles/static/browse.js
LeviButcher: I wanted browse.js to match the backend rendering of the paginator used on the recent page. On the backend it was less cumbersome to remove Next or Prev from the HTML conditionally then it was to toggle a disabled attribute since in the XHTML library there was no function for adding no attributes.

I wanted to do something like this: `a ! [href "URL", if hasNext then NoAttr else emptyAttr "disabled"] << "Next"`

That best way I could do it was to either to add a empty class or do disabled=True/False. Which browser defaults consider any presence of the disabled attribute whether True or False as disabled.

I can easily return the behavior though. :100:

src/Distribution/Server/Pages/Recent.hs
LeviButcher: Yup. Text-RSS doesn't have a way to pass multiple Links at the top level or give those links a "rel" attribute. I'd be more then willing to improve Text-RSS but that won't be quick.

datafiles/static/browse.js
LeviButcher: I added back in showing the disabled buttons with the most recent commit.

gbaz:
@ysangkok I'll of course take a closer look before acting on this, but want to know if from your standpoint you think the issues you raised have been sufficiently addressed?

ysangkok:
The paginator doesn't seem to work the same on the recent packages list, which I think it should. For example, I have one package in my index, and the browse page looks like:

image

Meanwhile, the recent packages listing looks like:

image

I don't think the ellipsis should be there. (I don't really care about the pluralization)

LeviButcher:
I've fixed the paginator issue @ysangkok described. :+1:

ysangkok:
A maximum page size like e.g. 200 should be enforced. Otherwise, it is really easy to overload the server. Especially since the parameter is received as a query string parameter, and therefore the overloading could be triggered simply by a link.

LeviButcher:
Implemented page size restriction. @ysangkok

Do you think the RecentPackages returning all packages is a performance issue? It does use the same caching as before although I wasn't for sure if I refactored it correctly. I can restrict it to only the past 48 worth or some x amount although that will restrict how many pages are possible.

ysangkok:
I think we should restrict that as well restrict that too, because it seems malicious if somebody is e.g. requesting page 50 of RecentPackages. If a legitimate user needs to do that, they're probably really trying to achieve something that would be better addressed by a new feature.

PR #1056, opened 2022-04-12
Fix some hlint warnings

jhrcek:
I'd like to contribute some more substantial fixes in the near future, but first please allow me to fix bunch of inconsequential hlint warnings, mostly just removing unused / unnecessary stuff


src/Distribution/Client/Mirror/Repo/Secure.hs
```diff
@@ -186,7 +186,7 @@ finalizeLocalMirror' cache repoRoot = (`runContT` return) $ do
     copyFileAtomic src dst = ContT $ \callback -> do
       let (dir, template) = splitFileName dst
       bracket (openBinaryTempFileWithDefaultPermissions dir template)
-              (\(temp, h) -> ignoreIOErrors (hClose h >> removeFile temp)) $
+              (\(temp, h) -> ignoreIOErrors (hClose h >> removeFile temp))
```
andreasabel, line 189: Here, rather than removing `$`, I would have removed the parentheses around the long term that follows.

src/Distribution/Server/Framework/BackupRestore.hs
```diff
@@ -108,9 +108,9 @@ abstractRestoreBackup putSt = go
   where
     go RestoreBackup {..} = AbstractRestoreBackup {
         abstractRestoreEntry = \store entry ->
-          liftM go    <$> (runRestore store $ restoreEntry entry)
+          fmap go    <$> runRestore store (restoreEntry entry)
```
andreasabel, line 111: In such situations, I use `do`:
```
fmap go <$> do runRestore store $ restoreEntry entry
```
(I don't think hlint knows that trick.)

andreasabel (approved):
99.9% of the suggestions are for the better! (For the rest, see my comments.)
You got my blessing!  (Assuming these rewrites are correct.)

src/Distribution/Server/Framework/BackupRestore.hs
jhrcek: Neat :smiley: applied

src/Distribution/Client/Mirror/Repo/Secure.hs
jhrcek: Makes sense, applied.

jhrcek:
Thank you for reviews, I consider this ready to merge.

gbaz:
I should add btw that the hlint changes are pretty minor as such, but since github was running them on every PR, cleaning up that noise really made development nicer. Thanks!

jhrcek:
> I should add btw that the hlint changes are pretty minor as such, but since github was running them on every PR, cleaning up that noise really made development nicer. Thanks!

Thanks for the feedback. I wasn't really sure if hlint fixes would be welcome by the maintainers, so I only fixed relatively conservative subset of warnings that I thought would be accepted by most people as obvious improvements.
Can you please tell me where exactly are you referring to by " cleaning up that noise"? Is it some view of hlint warnings in PR? Is it something produced by specific configuration of github actions/travis (if so can you point me to it so I understand where it's coming from)?

Anyway if you agree I will open followup PR, fixing more hlint warnings that make sense to be fixed.
In the end I could add an hlint config to ignore the rest of warnings that we agree we would want to ignore..

andreasabel:
> > I should add btw that the hlint changes are pretty minor as such, but since github was running them on every PR, cleaning up that noise really made development nicer. Thanks!
> 
> Thanks for the feedback. I wasn't really sure if hlint fixes would be welcome by the maintainers, so I only fixed relatively conservative subset of warnings that I thought would be accepted by most people as obvious improvements. 

Yes, I think this is very important.  I have some bad experiences with some external contributor pushing some hlint-suggested refactorings into the Agda code base, and some had quite negative effects on code readability.  This is why I looked at _each and every_ of your hlint-suggested changes in this PR and checked whether it aids readability.

> Can you please tell me what exactly are you referring to by " cleaning up that noise"? Is it some view of hlint warnings in PRs? Is it something produced by specific configuration of github actions/travis (if so can you point me to it so I understand where it's coming from)?

Er, we get a lot of noise from _GHC warnings_ (deprecated imports, incompleted pattern matching) but I do not think we have a hlint workflow afaik, so maybe @gbaz confused these two.

> Anyway if you agree I will open followup PR, fixing more hlint warnings that make sense to be fixed. In the end I could add an hlint config to ignore the rest of warnings that we agree we would want to ignore..

That sounds reasonable.  
Let me stress again that one should maintain a critical stance towards hlint since not all suggestions help readability.  E.g. introducing `foldr/l` often makes code harder to read.   Also, anecdotally, some hlint refactorings are unsound (break parsing or associate operators wrongly).

That said, I am happy to look at another round of likely-not-controversial hlint refactorings.
You did a very good job on this PR, @jhrcek !

gbaz:
right -- i confused the two because there are other codebases that do toss hlint warnings at me. whoops :-)

It would be nice to fix the ghc warnings for pattern matching as well I suppose, although doing it "right" would certainly take a bit of thought.

jhrcek:
> It would be nice to fix the ghc warnings for pattern matching as well I suppose, although doing it "right" would certainly take a bit of thought.

Those should be fairly easy to work around. I will look at fixing those tomorrow.

PR #1057, opened 2022-04-13
Fix some hlint warnings - part 2

jhrcek:
Second batch of hopefully non-controversial hlint fixes (as discussed in https://github.com/haskell/hackage-server/pull/1056#issuecomment-1097855098).

### DIFF
--- a/src/Distribution/Server/Features/BuildReports/BuildReports.hs
+++ b/src/Distribution/Server/Features/BuildReports/BuildReports.hs
@@ -89,8 +89,8 @@ data PkgBuildReports = PkgBuildReports {
     buildStatus :: !BuildStatus
 } deriving (Eq, Typeable, Show)
 
-data BuildReports = BuildReports {
-    reportsIndex :: !(Map.Map PackageId PkgBuildReports)
+newtype BuildReports = BuildReports {
+    reportsIndex :: Map.Map PackageId PkgBuildReports
 } deriving (Eq, Typeable, Show)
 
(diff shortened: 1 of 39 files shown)


src/Distribution/Server/Features/BuildReports/BuildReports.hs
```diff
@@ -89,8 +89,8 @@ data PkgBuildReports = PkgBuildReports {
     buildStatus :: !BuildStatus
 } deriving (Eq, Typeable, Show)
 
-data BuildReports = BuildReports {
-    reportsIndex :: !(Map.Map PackageId PkgBuildReports)
+newtype BuildReports = BuildReports {
```
jhrcek, line 92: Of all the "use newtype" suggestions I only applied those where the unrerlying field was strict, so as not to change strictness properties of any of the single-constructor single-field datatypes.

andreasabel (changes requested):
I would abstain from the `data -> newtype` conversions. (see my comment)
Everything else has my 100% approval!

src/Distribution/Server/Features/BuildReports/BuildReports.hs
jhrcek: Ok, makes sense. All the newtype changes reverted.

jhrcek:
@andreasabel friendly ping :smile_cat:

gbaz (approved):

andreasabel (commented):
Thanks for update!  Approved!

PR #1059, opened 2022-04-16
Fix incomplete-record-updates warnings

jhrcek:
Fixing some ghc warnings that show up in CI as discussed in https://github.com/haskell/hackage-server/pull/1056#issuecomment-1100179791

There are some issues in ghc with -fwarn-incomplete-record-updates giving false positives, e.g. 
https://gitlab.haskell.org/ghc/ghc/-/issues/5728
https://gitlab.haskell.org/ghc/ghc/-/issues/17783

Some of those should be fixed in recent versions of GHC (9+),
but we can avoid them even on GHC 8 by using RecordWildCards trick mentioned [here](https://gitlab.haskell.org/ghc/ghc/-/issues/5728#note_60108)
The other instances of this warning are fixed by restructuring the code a bit to make it more obvious to GHC that all cases are handled.

### DIFF
--- a/src/Distribution/Server/Framework/Auth.hs
+++ b/src/Distribution/Server/Framework/Auth.hs
@@ -428,26 +428,28 @@ data AuthError = NoAuthError
 
 authErrorResponse :: MonadIO m => RealmName -> AuthError -> m ErrorResponse
 authErrorResponse realm autherr = do
-    digestHeader   <- liftIO (headerDigestAuthChallenge realm)
-    return $! (toErrorResponse autherr) { errorHeaders = [digestHeader] }
-  where
-    toErrorResponse :: AuthError -> ErrorResponse
-    toErrorResponse NoAuthError =
-      ErrorResponse 401 [] "No authorization provided" []
+    digestHeader <- liftIO (headerDigestAuthChallenge realm)
+    let
+      toErrorResponse :: AuthError -> ErrorResponse
+      toErrorResponse ae = case ae of
+        NoAuthError ->
+          ErrorResponse 401 [digestHeader] "No authorization provided" []
 
-    toErrorResponse UnrecognizedAuthError =
-      ErrorResponse 400 [] "Authorization scheme not recognized" []
+        UnrecognizedAuthError ->
+          ErrorResponse 400 [digestHeader] "Authorization scheme not recognized" []
 
-    toErrorResponse InsecureAuthError =
-      ErrorResponse 400 [] "Authorization scheme not allowed over plain http"
-        [ MText $ "HTTP Basic and X-ApiKey authorization methods leak "
-               ++ "information when used over plain HTTP. Either use HTTPS "
-               ++ "or if you must use plain HTTP for authorised requests then "
-               ++ "use HTTP Digest authentication." ]
+        InsecureAuthError ->
+          ErrorResponse 400 [digestHeader] "Authorization scheme not allowed over plain http"
+            [ MText $ "HTTP Basic and X-ApiKey authorization methods leak "
+                   ++ "information when used over plain HTTP. Either use HTTPS "
+                   ++ "or if you must use plain HTTP for authorised requests then "
+                   ++ "use HTTP Digest authentication." ]
 
-    toErrorResponse BadApiKeyError =
-      ErrorResponse 401 [] "Bad auth token" []
+        BadApiKeyError ->
+          ErrorResponse 401 [digestHeader] "Bad auth token" []
 
-    -- we don't want to leak info for the other cases, so same message for them all:
-    toErrorResponse _ =
-      ErrorResponse 401 [] "Username or password incorrect" []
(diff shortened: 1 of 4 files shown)


src/Distribution/Server/Framework/HappstackUtils.hs
```diff
@@ -172,14 +172,14 @@ enableGZip compressParams modifyETag = do
...
-    gzipFilter r@Response{} =
+    gzipFilter Response{..} =
         chunked
       . setHeader "Content-Encoding" "gzip"
       . setHeader "Vary" "Accept-Encoding"
       . removeResponseHeader "Content-Length"
       . removeResponseHeader "Content-MD5"
       . alterResponseHeader "ETag" (map modifyETag)
-      $ r { rsBody = GZip.compressWith compressParams $ rsBody r }
+      $ Response { rsBody = GZip.compressWith compressParams rsBody, .. }
```
andreasabel, line 182: I find this version (with the magic dots `..` in the record update) harder to comprehend.  I consider this a fringe feature of GHC Haskell (I am seeing it for the first time now).
There might be some risks of unintentionally introducing bugs by shadowing one of the record fields introduced by the `..` pattern which is then picked up by the `..` update, is there?

I think that in the contrary we should strive to get rid of all the `..` magic and be more explicit about what is in scope.

So, I'd rather not change code to work around an incorrect GHC warning mechanism.  Rather switch off the warning.

src/Distribution/Server/Framework/HappstackUtils.hs
```diff
@@ -215,12 +215,15 @@ enableRange = do
...
     -- out the original length.
     rangeFilter :: (Int64, Int64) -> Response -> Response
-    rangeFilter (fr, to) r =
+    rangeFilter _ r@SendFile{} = r
+    rangeFilter (fr, to) r@Response{..} =
         setHeader "Content-Length" (show rangeLen)
       . setHeaderBS (BS.C8.pack "Content-Range") (contentRange fr to fullLen)
       . removeResponseHeader "Content-MD5"
-      $ r { rsBody = BS.Lazy.take rangeLen $ BS.Lazy.drop fr $ rsBody r
+      $ Response
```
andreasabel, line 223: Same here, I am opposed.

andreasabel (changes requested):
Thanks for your efforts!

As it stands, I am opposed to code changes just to work around false positives in a GHC warning mechanism.  If the warning is unsound, we should switch it off rather than change the  code (unless we can change the code without making any sacrifices, but I do not see it here).

So I am +1 on silencing the warning with `-Wno-...`, ideally in a per-file basis.
I am -1 on these code changes.

jhrcek:
It's not that -fwarn-incomplete-record-updates is unsound.
It actually found bunch of places that could potentially fail at runtime with incomplete pattern marches.
The only issue is that GHCs prior to version 9 can give this warning even for some cases where the failure condition provably cannot happen. I removed the usage of RecordWildCards from my PR. Let me know if you're against even this less invasive vefsion of the fixes and I'll close this PR.
Btw. any idea why CI is failing on 9.0.2?

andreasabel:
> Btw. any idea why CI is failing on 9.0.2?

Yes, this is mysterious.  I am suspecting that the package-db caching is unsound: https://github.com/haskell/hackage-server/blob/71bc80143cdd614de703b34c81e962754f2dc126/.github/workflows/ci.yml#L41-L47

I am pushing a new, idiomatic CI from
- #1061

src/Distribution/Server/Framework/Auth.hs
```diff
@@ -428,26 +428,28 @@ data AuthError = NoAuthError
...
-    toErrorResponse BadApiKeyError =
-      ErrorResponse 401 [] "Bad auth token" []
-
-    -- we don't want to leak info for the other cases, so same message for them all:
-    toErrorResponse _ =
-      ErrorResponse 401 [] "Username or password incorrect" []
+    digestHeader <- liftIO (headerDigestAuthChallenge realm)
+    let
+      toErrorResponse :: AuthError -> ErrorResponse
+      toErrorResponse ae = case ae of
```
andreasabel, line 434: I'd use `LambdaCase`:
```haskell
toErrorResponse = \case
```
which eliminates `ae` from the scope.

andreasabel (approved):

andreasabel:
I am ok with this modified PR.  Using `\case` would be a cosmetical improvement.
If you rebase this PR onto master, I hope the CI will succeed!

src/Distribution/Server/Framework/Auth.hs
jhrcek: Ok, I switched to using LambdaCase

jhrcek:
Thank you for updating CI, LambdaCase suggestion applied.

andreasabel:
Thanks, @jhrcek !

PR #1060, opened 2022-04-16
Add instructions for native deps installation on Fedora

jhrcek:
I don't insist on adding these instructions, but I had to install these native dependencies to be able to compile this project on Fedora.
Feel free to close this if you think the "installation instructions" section is getting to bloated.


andreasabel:
Thanks!
I am +1, as it is easier to ignore given information than to reconstruct information that is missing.

PR #1061, opened 2022-04-16
Auto-generate more complete CI via haskell-ci

andreasabel:
Atm. requires a patched version of `haskell-ci` 
- https://github.com/haskell-CI/haskell-ci/pull/582

to include the "haddock" step.

For now, this PR replaces 
- #1006

which instead patched the workflow produced  by `haskell-ci`.

Also: 
- fix missing `extra-source-files:` in `.cabal` file (detected by this CI (`haskell-ci.yml`)  but not the previous one (`ci.yml`))
- delete `.travis.yml`.


andreasabel:
The old CI can easily be restored by:
- deleting `.github/workflows/haskell-ci.yml`
- renaming `.github/workflows/ci.yml.deactivated` to `.../ci.yml`

However, the old CI likely handles the cache wrongly.

jneira:
I see the point of using haskell-ci but you could got a workflow for the three or two main os's with the half of code, but I guess you already know it
Also dependabot could upgrade actions automatically

andreasabel:
> I see the point of using haskell-ci but you could got a workflow for the three or two main os's with the half of code, but I guess you already know it Also dependabot could upgrade actions automatically

Ah, actually I don't know.  Do you have a pointer to a CI in that style, with some automatic upgrade mechanism?

jneira:
> Ah, actually I don't know. Do you have a pointer to a CI in that style, with some automatic upgrade mechanism?

dependabot opens prs to upgrade github actions, see https://github.com/haskell/haskell-language-server/pull/2823 for example
you can do it with haskell-ci but the updated config would be overwritten by the next regeneration

that is the main drawback of code generators, they dont compose (or compose hardly)

PR #1074, opened 2022-05-15
fix(server): email address validation

### LINKED ISSUE #1073: Cannot register new accounts
Hi,

When trying to register a new account on Hackage with a valid email address, you are redirected to the following error page:

_Problem with email address
Oops, that doesn't look like an email address._

Screenshot_20220515_164917

This looks like an error in the email address validation logic.

tochicool:
Fixes #1073

### DIFF
--- a/src/Distribution/Server/Util/Validators.hs
+++ b/src/Distribution/Server/Util/Validators.hs
@@ -30,11 +30,13 @@ guardValidLookingEmail str = either errBadEmail return $ do
   guard (T.all (not.isAngle) str) ?! "Please use just the email address, not \"name\" <person@example.com> style."
   where
     isAngle c = c == '<' || c == '>'
-    hasAtSomewhere =
-      let (before, after) = T.span (/= '@') str
-       in T.length before >= 1
-       && T.length after  >  1
-       && not ('@' `T.elem` after)
+    hasAtSomewhere = case T.span (/= '@') str of
+      (before, rest)
+        | Just (_, after) <- T.uncons rest ->
+          T.length before >= 1
+            && T.length after > 0
+            && not ('@' `T.elem` after)
+      _ -> False
 
 errBadUserName, errBadRealName, errBadEmail :: String -> ServerPartE a
 errBadUserName err = errBadRequest "Problem with login name" [MText err]


gbaz (approved):

gbaz:
Thanks!

ysangkok:
Hmm, had I broken this in #1047? I thought I just moved code around.

ysangkok:
Oh indeed I did, by trying to prevent more than one at-sign. I had misunderstood what `span` does:

```
Prelude Data.Text> Data.Text.span (/= '@') $ pack "a@b.com"
("a","@b.com")
```

Sorry.

src/Distribution/Server/Util/Validators.hs
```diff
@@ -30,11 +30,13 @@ guardValidLookingEmail str = either errBadEmail return $ do
...
-    hasAtSomewhere =
-      let (before, after) = T.span (/= '@') str
-       in T.length before >= 1
-       && T.length after  >  1
-       && not ('@' `T.elem` after)
+    hasAtSomewhere = case T.span (/= '@') str of
+      (before, rest)
+        | Just (_, after) <- T.uncons rest ->
+          T.length before >= 1
+            && T.length after > 0
```
andreasabel, line 37: What is the significance of using `>= 1` vs. `> 0`?  Why aren't both checks written as `not . T.null`?

andreasabel:
@ysangkok wrote:
> Sorry.

Better unit test than sorry. ;-)

More seriously, I am worried that there is no _beta_ deployment of Hackage.  I voiced this during one of our jitsi meetings.  I remember @gbaz saying that this is technically difficult.  In principle, this should not be difficult.  The database layer of Hackage needs to be separated from the UI layer, so that different UIs can connect to the same database.  Maybe the current architecture does not permit this.  Then we need to get some money to get the architecture modernized by some professionals like Well-Typed (who did the original architecture, AFAIK).

src/Distribution/Server/Util/Validators.hs
tochicool: Thanks, I didn't spot this.

`Text` is strict here so I don't think there's any functional difference. I didn't realise that `T.length` is O(n) so `not . T.null` is asymptotically better here, although we've already paid that cost with the guard that `T.length str <= 100` which would maybe be more efficient as `T.compareLength str 101 == LT`. We could consolidate the check further with ``((&&) `on` (not . T.null)) before after``.

andreasabel: +1.

> We could consolidate the check further with `` ((&&) `on` (not . T.null)) before after ``.

Mmh, this last suggestion seems just to make the code less readable...

andreasabel: I am following up on my suggestion in
- #1077

andreasabel:
> Better unit test than sorry. ;-)

Unit tests added in:
- #1077

Issue #1078, opened 2022-05-18
Should the Nix CI do caching


andreasabel:
@peterbecich You added a Nix CI:
https://github.com/haskell/hackage-server/blob/2d4f3f7fe3a5e38cb959eea139160caf273227ac/.github/workflows/nix-shell.yml#L16-L18

Does this cache between builds?

If not, would caching be possible, maybe along the lines of:
https://github.com/cachix/install-nix-action#how-do-i-add-a-binary-cache

I am asking because the Nix workflow takes at least 17min each time, whereas Haskell CI can get down to <10min (probably thanks to caching).
(Haskell CI takes long when it cannot restore from the cache, see e.g.: https://github.com/haskell/hackage-server/runs/6437475007?check_suite_focus=true#step:16:23)

peterbecich:
Good idea, working on it

It doesn't seem to produce an improvement to use https://github.com/cachix/install-nix-action#how-do-i-add-a-binary-cache

It would be an improvement to have `actions/cache@v3.0.2` working and cache `/nix/store` in Github; I'm trying to make this work: https://github.com/cachix/install-nix-action/issues/56#issuecomment-1030697681

The current errors are:
```
...
/usr/bin/tar: ../../../../../nix/store/01dww3kbj952cbbacnwbpzc4y3qr7clc-glibc-2.34-115-bin/sbin: 
Cannot create symlink to ‘bin’: File exists
/usr/bin/tar: ../../../../../nix/store/01dww3kbj952cbbacnwbpzc4y3qr7clc-glibc-2.34-115-bin/bin/xtrace: 
Cannot open: File exists
/usr/bin/tar: ../../../../../nix/store/01dww3kbj952cbbacnwbpzc4y3qr7clc-glibc-2.34-115-bin/bin/zic: 
Cannot open: File exists
/usr/bin/tar: ../../../../../nix/store/01dww3kbj952cbbacnwbpzc4y3qr7clc-glibc-2.34-115-bin/bin/getent: 
Cannot open: File exists
/usr/bin/tar: ../../../../../nix/store/01dww3kbj952cbbacnwbpzc4y3qr7clc-glibc-2.34-115-bin/bin/ldd: 
Cannot open: File exists
...
```

peterbecich:
As far as I can tell, the Cachix GitHub Action (https://github.com/cachix/install-nix-action) is designed to only cache on Cachix, and is not compatible with https://github.com/actions/cache

[...]

PR #1086, opened 2022-06-26
attempt to speed up GitHub Action for Nix Shell

peterbecich:
https://github.com/haskell/hackage-server/issues/1078

Unsure if this will work.
It appears `cabal update` is necessary but makes the cache unusable. Maybe using a specific date will avoid this:
https://github.com/haskell/hackage-server/blob/9142bb53a3fdd206504a4dddca1a7ad03989f0e9/.github/workflows/test-nix-shell.sh#L3

### DIFF
--- a/.github/workflows/test-nix-shell.sh
+++ b/.github/workflows/test-nix-shell.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+cabal update hackage.haskell.org,2022-08-27T00:00:00Z
+cabal build all --enable-tests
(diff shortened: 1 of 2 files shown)


peterbecich:
@andreasabel please review. I'm not sure if it will make an improvement until it is merged. I think PRs are not allowed to push to Cachix.

.github/workflows/test-nix-shell.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+cabal update hackage.haskell.org,2022-08-27T00:00:00Z
```
andreasabel, line 3: Does this timestamp have to be updated manually frequently?  Should there be instructions when to do so?

andreasabel:
@peterbecich Thanks! I merged it for testing, please evaluate results.

PR #1087, opened 2022-06-27
Fix analyticsPixel template variable name

alexbiehl:
This PR addressess @gbaz's comment 

> I merged in preperation for a redeploy, but noticed in testing that the list of pixels already created did not show up properly 
> in the package page. (It does show up in the page that lists all pixels for all packages for a user). So I'm going to comment > out links to this functionality for now in the release. A patch fixing that up would be welcome.

in https://github.com/haskell/hackage-server/pull/1042#issuecomment-1126827553


fgaz:
Big -1 to this and to the whole tracking pixel feature!

Google Analytics was just removed in #1097 because of privacy and legality concerns. While this is more limited from a technical pov, legally it's still the same since tracking pixels can collect personal data such as IP addresses. Worse than that, this puts control over the users' data in the hands of maintainers (who would control the tracking settings), while putting the _responsibility_ in the hands of hackage operators and haskell.org (data controller)! That's a huge burden and has too much potential to be abused!

PR #1089, opened 2022-06-29
Allow hashable-1.4 and text-2.0

Bodigrim:


gbaz (approved):

andreasabel:
Note that while `hashable-1.4` is picked up by CI, `text-2.0` isn't, so the latter is untested afaict.

Issue #986, opened 2021-10-24
Implement PackageRank from lib.rs


gbaz:
A very impressive amount of research and thought here: https://github.com/rust-lang/rfcs/blob/master/text/1824-crates.io-default-ranking.md

It would be useful to add this as an explicit metric to sort by. 🤔

Bodigrim:
It seems lib.rs essentially orders by number of recent downloads, or am I missing something? It should be straightforward to add alongside with existing total downloads.

gbaz:
Crates.io orders by recent downloads, which we have (though inaccurate because of cdn and can do). I should have really linked this, which explains lib.rs uses a much more sophisticated algorithm: https://lib.rs/about

When I made this ticket, I was under the misapprehension the crates ticket somehow contained this algo, but I don't think it does (just related research). The ranking itself can be figured out from the code here: https://gitlab.com/crates.rs/crates.rs/-/blob/main/ranking/src/lib_ranking.rs

PR #1091, opened 2022-07-01
PageRank

kubaneko:
This PR is for implementing #986 as an GSoC project. Mentored by @gbaz .

I will correct the error I made in the change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# hackage-server
+# Hackage-server
 
 [![Build Status](https://travis-ci.org/haskell/hackage-server.png?branch=master)](https://travis-ci.org/haskell/hackage-server)
 [![Build status](https://github.com/haskell/hackage-server/actions/workflows/haskell-ci.yml/badge.svg)](https://github.com/haskell/hackage-server/actions/workflows/haskell-ci.yml)


src/Distribution/Server/Features/Browse/Parsers.hs
```diff
@@ -113,6 +113,7 @@ filterWith allowNot = do
         "maintainer:" -> MaintainerFilter <$> wordWoSpaceOrParens
         "deprecated:" -> DeprecatedFilter <$> deprecatedOption
         "distro:" -> DistroFilter <$> wordWoSpaceOrParens
+        "packageRank" -> DownloadsFilter <$> opAndSndParam decimal
```
ysangkok, line 116: The package rank should have it's own constructor, it makes little sense to use DownloadsFilter

datafiles/static/browse.js
```diff
@@ -134,6 +134,7 @@ const replaceRows = (response) => {
     tr.appendChild(createTags(row.tags));
     tr.appendChild(createLastUpload(row.lastUpload));
     tr.appendChild(createMaintainers(row.maintainers));
+    tr.appendChild(createSimpleText(row.packageRank));
```
ysangkok, line 137: This also needs a table header field in `datafiles/templates/Html/browse.html.st`

datafiles/templates/Html/noscript-search-form.html.st
```diff
@@ -23,6 +23,7 @@
         <option $if(tags)$        selected $endif$ value=tags>tags</option>
         <option $if(lastUpload)$  selected $endif$ value=lastUpload>date of last upload</option>
         <option $if(maintainers)$ selected $endif$ value=maintainers>maintainers</option>
+        <option $if(packageRank)$ selected $endif$ value=packageRank>packageRank</option>
```
ysangkok, line 26: the text inside the option shouldn't be camelCase, since it is the human readable description of the column, shown in a drop-down

src/Distribution/Server/Features/Browse/Parsers.hs
kubaneko: For now I undid the changes to Parsers.hs since they do not make sense and are not necessary.

datafiles/templates/Html/noscript-search-form.html.st
kubaneko: Thank you for the fixes. I hope 7304a05 resolves the last two.

src/Distribution/Server/Features/PackageRank.hs
```diff
@@ -0,0 +1,281 @@
...
+  pkgId = package pkg
+  lastUploads =
+    sortBy (flip compare)
+      $  (fst . pkgOriginalUploadInfo <$> pkgs)
+      ++ (fst . pkgLatestUploadInfo <$> pkgs)
+  versionList :: [Version]
+  versionList = sortBy (flip compare)
+    $ map (pkgVersion . package . packageDescription) (pkgDesc <$> pkgs)
+  packageEntr = do
+    tarB <- packageTarball tarCache . head $ pkgs
```
ysangkok, line 134: Avoidable partial function

src/Distribution/Server/Features/PackageRank.hs
```diff
@@ -0,0 +1,281 @@
...
+  -> IO Scorer
+
+rankIO vers recentDownloads maintainers docs env tarCache pkgs = do
+  temp  <- temporalScore pkg lastUploads versionList recentDownloads
+  versS <- versionScore versionList vers lastUploads pkg
+  codeS <- codeScore documentLines srcLines
+  return (temp <> versS <> codeS <> authorScore maintainers pkg)
+
+ where
+  pkg   = packageDescription <$> pkgDesc $ last pkgs
```
ysangkok, line 124: Avoidable partial function

src/Distribution/Server/Features/PackageList.hs
```diff
@@ -243,19 +255,22 @@ listFeature CoreFeature{..}
...
-        items <- mapM (constructItem . last) $ PackageIndex.allPackagesByName index
+        items <- mapM constructItem $ PackageIndex.allPackagesByName index
         return $ Map.fromList items
 
-    constructItem :: PkgInfo -> IO (PackageName, PackageItem)
-    constructItem pkg = do
+    constructItem :: [PkgInfo] -> IO (PackageName, PackageItem)
+    constructItem pkgs = do
         let pkgname = packageName pkg
+            pkg = last pkgs
```
ysangkok, line 264: Avoidable partial function

src/Distribution/Server/Features/PackageRank.hs
```diff
@@ -0,0 +1,281 @@
...
+  -> DocumentationFeature
+  -> TarIndexCacheFeature
+  -> ServerEnv
+  -> [PkgInfo]
+  -> IO Float
+rankPackage versions recentDownloads maintainers docs tarCache env pkgs =
+  total
+    .   (<>) (rankPackagePage pkgD)
+    <$> rankIO versions recentDownloads maintainers docs env tarCache pkgs
+  where pkgD = packageDescription $ pkgDesc $ last pkgs
```
ysangkok, line 281: Avoidable partial function

src/Distribution/Server/Features/PackageList.hs
```diff
@@ -222,7 +234,7 @@ listFeature CoreFeature{..}
                 let pkgs = PackageIndex.lookupPackageName index pkgname
                 case pkgs of
```
ysangkok, line 235: this can case on `Data.NonEmpty.nonEmpty` instead. Then the `Just nonEmptyList` case can pass that list on to `constructItem`

ysangkok (commented):
I think you can avoid these partial functions by using NonEmpty. See comments.

kubaneko:
@ysangkok

Thank you for the comments

In my last two commits I fixed all the partial functions in PackageRank.

The partial function in PackageList is in my opinion better left as is since it was that way even before my changes and on line 256 using NonEmpty would lead to Maybe PackageIndex. Which is not something that is wanted?

src/Distribution/Server/Features/PackageRank.hs
```diff
@@ -0,0 +1,289 @@
...
+  versionList :: [Version]
+  versionList = sortBy (flip compare)
+    $ map (pkgVersion . package . packageDescription) (pkgDesc <$> pkgs)
+  packageEntr = do
+    tarB <- packageTarball tarCache pkgI
+    return
+      $   (\(path, _, index) -> (path, ) <$> T.lookup index path)
+      =<< rightToMaybe tarB
+  rightToMaybe (Right a) = Just a
+  rightToMaybe (Left  _) = Nothing
```
ysangkok, line 141: Can't you take advantage of the fact that there is MonadFail Maybe? It also doesn't make sense to me to mix do-notation and calls to bind (`=<<`). I'd almost always prefer do notation since it is more readable to have identifiers.

I think the code is doing something like this, which I think is a lot simpler:

```
packageEntr = do
    Right (path, _, index) <- packageTarball tarCache pkgI
    foobar <- T.lookup index path
    pure (path, foobar)
```

kubaneko: This part of the code does not actually work so when I fix it I will format it nicer. I was not aware of MonadFail this will probably make a lot of things simpler.

kubaneko:
At this moment PackageRank should be structure-wise complete. Some parameters will need to be adjusted and there may be some bugs.

Caching some data (ex. freshness for dependencies would require parsing all their PkgInfo (not included)) is missing.

UI is a bit arbitrary so that might also change.

kubaneko:
image

UI at the moment @ysangkok

nikita-volkov:
What's blocking us from having this merged? Looks like a great feature to have.

kubaneko:
The indirect dependencies feature was merged and it would be nice to use it here, adding a cache for the package values would be good and the parameters are somewhat arbitrary at the moment.

I plan to get this into mergeable state in June.

ysangkok:
@kubaneko Good, that's great to hear.

kubaneko:
I rebased and added some version of #1082 to PackageRank. I tested this by mirroring 600 package from hackage and the numbers it gives look reasonable. 

The performance implications are unclear to me, since I do not have the download to mirror the whole hackage but it was not too bad (I am unsure if adding indirect dependencies is a problem).

There is definitely still work that can be done:
- in freshnessScore, freshness of dependencies should be taken into account
  - this would probably require a cache
- cache would also be a good idea since Package rank does do some computation
- tuning the parameters to Haskell would be nice
- refactoring the code a bit would help

Currently I would like if someone with a hackage mirror would try what the slowdown is. If it is not noticable than I propose to merge this otherwise I will get to making the cache.

Any more comments/suggestions are welcome

ysangkok (changes requested):

Issue #668, opened 2018-02-05
Package page should give links to last working docs when new docs aren't built?


gbaz:
From reddit: "It would be better to maybe have some copy around where the links are that says "The documentation for this version isn't available; the last version with documentation available is 0.0.1.0""

kanishka-azimi:
I do this all the time, just surfing backwards in versions until I get a haddock. Might be simpler to implement a small indicator next to each version indicating haddock present or not.

Issue #1095, opened 2022-07-07
/docs redirect doesn't take into account deprecated packages


ndmitchell:
In Hoogle, if you search for `FunBind` the first result links to https://hackage.haskell.org/package/ghc/docs/Language-Haskell-Syntax-Binds.html#v:FunBind. The URL https://hackage.haskell.org/package/ghc takes me to GHC 9.2.3, which is the second-latest version, as the subsequent ghc-9.2.3.20220620 has been marked deprecated. However, https://hackage.haskell.org/package/ghc/docs redirects to https://hackage.haskell.org/package/ghc-9.2.3.20220620/docs/ - the deprecated version, which in this case has no documentation.

It would be good if /docs shared the same logic as the non-docs link.

PR #1096, opened 2022-07-08
Link to last working doc

AliasQli:
This pull request implements issue #668 as an GSoC project, mentored by @gbaz. It adds an indicator for each package version with a working doc. It looks like this, where version `0.1.0` has doc but `0.1.1` not:

Screenshot_20220708_164253

Actually I'm not very satisfied with the UI and am expecting a better idea.

This is my first time contributing here, so if I did something wrong in code style or structure please point it out.

### DIFF
--- a/datafiles/templates/Html/package-page.html.st
+++ b/datafiles/templates/Html/package-page.html.st
@@ -114,9 +114,6 @@
             <tr>
               <th>Versions <span style="font-weight:normal;font-size: small;">[<a href="/package/$package.name$.rss">RSS</a>]</span></th>
               <td>$versions$</td>
-              $if(! hasDocs)$
-                <td class="word-wrap">$lastDoc$</td>
-              $endif$
             </tr>
 
             $if(package.optional.hasChangelog)$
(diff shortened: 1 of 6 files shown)


AliasQli:
Now I changed the UI and this time it's much better. When the current version doesn't have a doc built, it tries to find the greatest version with a working doc that is not depcrated. Screensoht:

gbaz:
I think that find code is correct. However, I think the link should not be in the versions listing.

Rather, it should be either above or below the modules listing in the main pane, not the sidebar. (As the modules listing is where people would click through to get the docs, should they exist:

image

Additionally, maybe the redirect for docs, as described in this issue, could pick last working docs: https://github.com/haskell/hackage-server/issues/1095

AliasQli:
The described redirect for docs is added. Now `DocumentationFeature` depend on `VersionsFeature`, since it has to find out the package status. Unpreferred versions are also excluded from redirection. Will work on UI tomorrow.

AliasQli:
Screenshot:

gbaz:
was accidentally included in https://github.com/haskell/hackage-server/pull/1116 and hence already merged

PR #1097, opened 2022-07-10
Revert "add GA tag"

fgaz:
This reverts commit 95826dfa47396d223ad58df91c53ce56fdb283a1.

The use of Google Analytics [has been declared unlawful in three EU countries](https://noyb.eu/en/update-further-eu-dpa-orders-stop-google-analytics) already, with probably more to follow.
Keeping it there is a huge liability.


gbaz:
I discussed this with fgaz and at this point I think that the very marginal benefit we get out of GA or any other analytics package is not worth the hassle of increasing rulings that make doing such things confusing and difficult.

Issue #637, opened 2017-10-22
Link to Maintainers from each package's hackage page


ivanperez-keera:
Each package has a maintainer, declared in the cabal file (e.g. [1]). Each package has a list of maintainers authorised to make changes to it on hackage (e.g. [2]).

It would be a nice small addition if there was a link from the former to the latter (from [1] to [2]).

[1] https://hackage.haskell.org/package/Yampa/
[2] https://hackage.haskell.org/package/Yampa/maintainers/

EDIT: clarification.

hvr:
@ivanperez-keera can you maybe create a mockup of the UI? I've had considered this too, but I couldn't come up with a UI design that wouldn't be somewhat weird, as the maintainer-group and the .cabal declared maintainer(s) have a bit of an impedance mismatch, being in a different format/namespace.

ivanperez-keera:
@hvr What I had in mind was super-simple. For example:
![screenshot from 2017-10-23 01-09-04](https://user-images.githubusercontent.com/1162320/31867396-202c89ae-b78f-11e7-9a5f-1df797f111eb.png)

hvr:
@ivanperez-keera ok, that's easy enough; and you see no problem if the .cabal maintainer field has completely diverged from the hackage-maintainer-group?

ivanperez-keera:
- Sure, it may be a bit confusing that they can diverge. And those being different is not always an indication of a problem. When both are real persons, then a reasonable expectation would be that the intersection of both not be empty. However, maintainers may not be real persons. For example, the "cabal" maintainer email might be a mailing list. And the "hackage" maintainer account might correspond to Travis CI. 

   A possibility to hint that this could happen would be to use a different word instead of details, such as 'official list', 'all', 'list' or 'hackage maintainers'. 

- The other, maybe less confusing option, may be to add it at the bottom, in the current _Maintainer's corner_. It might require renaming the section so that it fits semantically.

gbaz:
I think the bottom is a good idea. That's where I'd expect it, actually.

gbaz:
Note -- there's a little work that needs to be done to thread the data through to where it needs to go, as the maintainers group isn't normally available to that page directly.

andreasabel:
@ivanperez-keera : Are you happy with today's status quo? 
[...]

Issue #953, opened 2021-07-19
maintainer group should be explicitly listed on package page


gbaz:
its a bit inconvenient to only be able to click through to the user page of the last uploader, and not all maintainers (and also harder to discover).

phadej:
There are plenty of packages, with "honorary past" maintainers, e.g. https://hackage.haskell.org/package/aeson/maintainers/

I don't have any need to remove others from that list, but OTOH having Bryan mentioned on the aeson package won't serve any good.

There is `maintainer: ` field in the `.cabal` metadata (which can be updated via revision). How to avoid confusion between that and *people with upload rights* on Hackage?

i don't think this is good first issue as the spec is not clear.

andreasabel:
Maybe #918 would be enough for now...

gbaz:
closing as dup of https://github.com/haskell/hackage-server/issues/637 (which should be resolved by gsoc work)

PR #1098, opened 2022-07-10
List maintainers on package page

AliasQli:
This pull request implements issue #637 as a gsoc project mentored by @gbaz. It displays hackage maintainers in Maintainer's Corner, on both package and candidate pages. Looks like this:

### DIFF
--- a/src/Distribution/Server/Pages/Group.hs
+++ b/src/Distribution/Server/Pages/Group.hs
@@ -70,11 +70,11 @@ removeUser uname uri =
     ]
 
 listGroup :: [Users.UserName] -> Maybe String -> Html
-listGroup [] _ = p << "No member exist presently"
+listGroup [] _ = p << "No current members of group"
 listGroup users muri = unordList (map (displayName muri) users)
 
 listGroupCompact :: [Users.UserName] -> Html
-listGroupCompact [] = toHtml "No member exist presently"
+listGroupCompact [] = toHtml "No current members of group"
 listGroupCompact users = foldr1 (\a b -> a +++ ", " +++ b) (map (displayName Nothing) users)
 
 displayName :: Maybe String -> Users.UserName -> Html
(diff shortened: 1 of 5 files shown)


src/Distribution/Server/Pages/Group.hs
```diff
@@ -73,3 +74,8 @@ listGroup [] _ = p << "No member exist presently"
 listGroup users muri = unordList (map displayName users)
   where displayName uname = (anchor ! [href $ "/user/" ++ display uname] << display uname) +++
                             maybe [] (removeUser uname) muri
+
+listGroupCompact :: [Users.UserName] -> Html
+listGroupCompact [] = toHtml "No member exist presently"
+listGroupCompact users = foldr1 (\a b -> a +++ ", " +++ b) (map displayName users)
+  where displayName uname = (anchor ! [href $ "/user/" ++ display uname] << display uname)
```
andreasabel, line 81: The last line has copy-and-paste code from above.  Please factor out the common code.

andreasabel (changes requested):
Some changes requested (see above).

AliasQli:
> Is _Maintainers_ identical to _package maintainers_?

Yes. However, I don't understand what you mean by "the same terminology". Did you mean moving these links to the *Maintainers* section in the right?

gbaz:
As Alias says, the "maintainer" field on the right is a free text field in the cabal file. This "maintainers" is the listing of actual hackage accounts with maintainer rights for the package. So I think it should live where it does, and not on the right. We could rename this new one to "hackage maintainers" or "hackage maintainers group" or something more clarifying...

andreasabel:
All I meant is:
```
MAINTAINERS (linking to the hackage maintainers pages): Alice, Bob, Carlo
For MAINTAINERS and hackage trustees
* edit package information
```
Here, replace MAINTAINERS by either "maintainers" or "package maintainers", but use the same wording; otherwise one could think that different collections of individuals are meant there.
Put the link on the first occurrence of MAINTAINERS rather than on the second.

andreasabel (approved):

AliasQli:
image
Now this looks fine.

andreasabel:
Closes #637. Closes #953.

Issue #961, opened 2021-08-05
Tests Failed without any logs


konsumlamm:
I recently updated my package and the [build log](https://hackage.haskell.org/package/rrb-vector-0.1.1.0/reports/1) says "Tests Failed" without showing any log of what exactly went wrong. The test suite passes locally for me (using the same GHC version).

brandonchinn178:
Bump. Same for my package: https://hackage.haskell.org/package/tasty-autocollect-0.2.0.0/reports/1

konsumlamm:
Now at least there is a [log](https://hackage.haskell.org/package/rrb-vector-0.2.0.0/reports/1), but unfortunately that isn't very useful, since it doesn't show the test suite output.

PR #1100, opened 2022-07-13
Add test log display

AliasQli:
This pr implements #961 as a gsoc project mentored by @gbaz.


AliasQli:
The hackage-build exe creates a folder and puts all cabal files within, but in `parseRepositoryIndices` it uses the user's cabal folder. This causes great difficulty when working with a local hackage server, and I think this is a bug. The second commit fixes it. However, if this is a deliberate design, please revert the commit.

gbaz:
Also added a fix to read from 00-index.tar if 01-index.tar is not available. Still not great, eesh.

PR #1114, opened 2022-07-24
Fix #1105: change order of markdown parsers to allow pipe characters in lists

### LINKED ISSUE #1105: Weird markdown dash list rendering
Blocked by:
- https://github.com/jgm/commonmark-hs/issues/95

Issue description:
Screenshot 2022-07-18 at 20 29 39

Compare the first and second dash list.  This is the source:
```md
### 1.3.2

_2022-07-18, Andreas Abel_

- Export `Text.Regex.TDFA.Pattern.decode{PatternSet|CharacterClass}`
  ([#16](https://github.com/haskell-hvr/regex-tdfa/issues/16))
- Extend and correct docs for 'Pattern' module
- Tested with GHC 7.4 - 9.4

### 1.3.1.5

_2022-07-18, Andreas Abel_

- Allow dash (`-`) as start of a range, e.g. `[--z]`
  ([#1](https://github.com/haskell-hvr/regex-tdfa/issues/1),
  [#45](https://github.com/haskell-hvr/regex-tdfa/pull/45))
- Tested with GHC 7.4 - 9.4
```
Could be an upstream issue in `commonmark`, need to test there.
UPDATE: At least the vanilla [`commonmark` parser](https://github.com/jgm/commonmark-hs/blob/714a1b651a5a6dc0fa9e0a2eba9848ca898be383/commonmark/README.md?plain=1#L52-L61) treats this correctly.

andreasabel:
Fix #1105: change order of markdown parsers to allow pipes in lists

Workaround suggested at:
- https://github.com/jgm/commonmark-hs/issues/95

### DIFF
--- a/src/Distribution/Server/Util/Markdown.hs
+++ b/src/Distribution/Server/Util/Markdown.hs
@@ -120,6 +120,30 @@ adjustRelativeLink url
 -- <p>Published to <a href="http://hackage.haskell.org/foo3/bar">http://hackage.haskell.org/foo3/bar</a>.</p>
 -- <BLANKLINE>
 --
+-- >>> renderMarkdown "test" "Issue #1105:\n- pipes\n- like `a|b`\n- should be allowed in lists"
+-- <p>Issue #1105:</p>
+-- <ul>
+-- <li>pipes
+-- </li>
+-- <li>like <code>a|b</code>
+-- </li>
+-- <li>should be allowed in lists
+-- </li>
+-- </ul>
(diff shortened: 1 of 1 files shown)


jgm:
A better description would be: "to allow pipe characters in lists"; without this change, pipe tables in lists were possible, it was lines that could have been the first line of a pipe table, but were not, that caused problems.

andreasabel:
Let's try this out, I suppose.  However, we wouldn't see any feedback until the next deployment...

PR #1117, opened 2022-08-07
Ask new uploaders to link to their packages

endgame:
New accounts are telling us what package they'd like to upload, which is good, but then they're emailing us cabal sdists, which is less good.

### DIFF
--- a/datafiles/templates/UserSignupReset/SignupConfirmation.email.st
+++ b/datafiles/templates/UserSignupReset/SignupConfirmation.email.st
@@ -7,10 +7,11 @@ this link:
   $confirmlink$
 
 After your account is created, you will need to email the Hackage
-Trustees at hackage-trustees@haskell.org requesting to be added to the
+Trustees at hackage-trustees@haskell.org requesting addition to the
 uploader group. In this email, please include your Hackage username
-and a package that you'd like to upload. (This measure is
-unfortunately necessary to prevent spam accounts.)
+and a link to the package you'd like to upload (on a host like gitlab
+or github) that you'd like to upload. (This measure is unfortunately
+necessary to prevent spam accounts.)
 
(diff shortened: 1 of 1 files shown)


gbaz (approved):

endgame:
CI breakage doesn't look like my fault, at a glance.

endgame:
Also the text is currently incoherent. I'll need to fix that before merge.

endgame:
Okay, I think that reads better.

datafiles/templates/UserSignupReset/SignupConfirmation.email.st
```diff
@@ -7,10 +7,11 @@ this link:
   $confirmlink$
 
 After your account is created, you will need to email the Hackage
-Trustees at hackage-trustees@haskell.org requesting to be added to the
+Trustees at hackage-trustees@haskell.org requesting addition to the
 uploader group. In this email, please include your Hackage username
-and a package that you'd like to upload. (This measure is
-unfortunately necessary to prevent spam accounts.)
+and a link to the package you'd like to upload (on a host like gitlab
+or github). This measure is unfortunately necessary to prevent spam
```
vaibhavsagar, line 13: Do we want to add an "if possible" here?

vaibhavsagar (approved):

datafiles/templates/UserSignupReset/SignupConfirmation.email.st
endgame: I don't think so, it's table stakes for an opensource project and if it's under a hackage-compatible licence anyone could stand one up.

endgame:
What needs to happen to get this merged?

gbaz:
I'll merge it. No promises on when the next deploy will be however.

PR #1126, opened 2022-08-19
Upgrade to haddock-library-1.11.0

### LINKED ISSUE #1125: Can't build with updated dependency
The dependency `haddock-library` updated from 1.10.0 to 1.11.0, and made an uncompatible change. The type signature of `DocMarkUpH`'s field `markupOrderedList` changed from `[a] -> a` to `[(Int, a)] -> a`. The version bound for `haddock-library` is `> 1.7 && < 2`, and as a result, the repo can't build with the latest dependencies.

Should we tighten the version bound for this library?

AliasQli:
This pr resolves #1125.


AliasQli:
It seems that CI's ghcs have the version fixed for the `process` package, and none of them is in the version bound of `Cabal-3.8.1.0`: `process (>=1.2.1.0 && <1.7)`. Don't know what to do.

AliasQli:
Maybe it's too early to move to Cabal 3.8; as for now, staying at Cabal 3.6 should be fine.

gbaz:
I'm not sure why the CI's being fixed against a given _process_ means that a newer one is uninstallable? Thoughts @ysangkok ?

ysangkok:
@AliasQli All of the recent GHCs have `process` versions that are in the range you provided, see [version history on the ghc gitlab](https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/libraries/version-history).

Could you provide more details on your reasoning? If you provide all the details it will be easier to understand what is not behaving as you expect.

@gbaz Why do you say that CI is fixed against a certain process version?

All I know is that cabal treats boot libraries in a special way, and there is a bug regarding them. For example, if I add `mtl-2.3.1-rc1` as a source-repository-package, but I don't set `--constraint=transformers>=0.6'`, I think it will try to build with transformers-0.5 (boot version). So when you get weird solver issues, I suggest providing more constraints with `==`. That way, you are forced to look up the expected/newest version on Hackage, and you may discover that the solver somehow picked a super old release of some package (old version often have bad bounds). I think I hit the issue about constraints not being applied yesterday also, where cabal was [not discovering that the boot library version of `containers` was too new](https://github.com/goldfirere/th-desugar/issues/157#issuecomment-1221613230) for `th-lift-instances`. But if I provided the constraint on containers, it did discover it.

AliasQli:
I don't understand the mechanism exactly, but this is what I've tested and found out. I have
```
build-depends:
      process                == 1.6.14.0
    , ghc
```
Cabal solved the version constraints locally but fails on CI, and the error message suggests `process` version is fixed: see https://github.com/AliasQli/hackage-server/actions/runs/2910726317.

src/Distribution/Server/Features/Search/ExtractDescriptionTerms.hs
```diff
@@ -78,7 +78,7 @@ termsMarkup = Markup {
   markupBold          = id,
   markupMonospaced    = \s -> if length s > 1 then [] else s,
   markupUnorderedList = concat,
-  markupOrderedList   = concat,
+  markupOrderedList   = concat . map snd,
```
andreasabel, line 81: This won't work for `haddock-library-1.10.0`.  You will also have to bump the lower bound of `haddock-library` to `>= 1.11.0` if you want to sort this out without CPP directives like `#if MIN_VERSION_haddock_library(1,11,0)`.

hackage-server.cabal
```diff
@@ -118,6 +118,7 @@ common defaults
   build-depends:
     , aeson                 ^>= 2.0.3.0
     , Cabal                 ^>= 3.6.3.0
+    , Cabal-syntax          ^>= 3.6
```
andreasabel, line 121: I have added this line in PR
- #1131

andreasabel (changes requested):
Need to make sure it works with all versions of `haddock-library` in range.

andreasabel:
@AliasQli wrote:
> the error message suggests `process` version is fixed

Indeed, Hackage-CI fixes the packages that are shipped with GHC.  If you want to relax this by adding `-process` to this line: https://github.com/haskell/hackage-server/blob/2377900828ba1d3086cbb01f15f45573ae05edfa/cabal.haskell-ci#L3
And then run `hackage-cli regenerate`. 

However, I advise you against pinning down a specific version of `process`.  I do not see the need for it.  It is better left at the versions shipped with the individual GHC versions.

andreasabel:
I left some comments at: https://github.com/haskell/hackage-server/commit/f730854464a19a179a7a9300756f57676510e343

This PR should be rebased onto latest `master`.

AliasQli:
Oops, I forgot to tighten the version bounds.

@andreasabel 

> And then run `hackage-cli regenerate`.

I suppose you mean `haskell-ci regenerate`. I've tested it, and  it works only when the project doesn't depend on the `ghc` package. When we have `ghc` in the dependecies (which is our case), CI still stick to the installed version of `process`.

andreasabel:
> I suppose you mean `haskell-ci regenerate`.

Yes, sorry.  These two tools have names too similar for my brain to not mix them up constantly... ;-)

> When we have `ghc` in the depende[n]cies (which is our case), CI still stick to the installed version of `process`.

Interesting.  I was going to say that `ghc` might bring further constraints on `process`, but I cannot find any evidence for this.

andreasabel (approved):
Thanks!

andreasabel:
This PR has conflicts with `master` that need to be resolved.

AliasQli:
I didn't see any conflicts. Could you please point them out to me?

hackage-server.cabal
```diff
@@ -111,8 +111,8 @@ common defaults
     , process                >= 1.6   && < 1.7
     , text                  ^>= 1.2.5.0 || ^>= 2.0
     , time                   >= 1.9   && < 1.13
-    , transformers           >= 0.5   && < 0.7
-    , unix                   >= 2.7   && < 2.9
+    , transformers           >= 0.5   && < 0.6
+    , unix                   >= 2.7   && < 2.8
```
andreasabel, line 115: Please don't include untested major versions bump.

andreasabel (changes requested):

hackage-server.cabal
AliasQli: `transformers` 0.5 is tested while 0.6 is not, and so is `unix`. So I tightened the version bounds. What's the problem?

AliasQli: I did it because you mentioned it [here](https://github.com/haskell/hackage-server/commit/f730854464a19a179a7a9300756f57676510e343#r82008276). Did I misunderstand what you mean?

andreasabel: Sorry, I confused this (I wasn't expecting these bounds already on master).  All is well.

PR #1137, opened 2022-09-23
Add support for 64-bit RISC-V architecture

felixonmars:
This helps some projects to keep support for this architecture (for example: https://github.com/haskell-primitive/primitive-unaligned/commit/2a8ec3167e9cabb520884a7e963ec6bc733983d6)


gbaz:
Doesn't appear to be a valid arch https://hackage.haskell.org/package/Cabal-syntax-3.8.1.0/docs/Distribution-System.html#t:Arch ?

felixonmars:
Ah... Looks like it should be added there as well.

Bodigrim:
This will be unblocked once https://github.com/haskell/hackage-server/pull/1322 is merged.

andreasabel:
Subsumed by:
- #1322

PR #1141, opened 2022-10-25
Build with Cabal-3.8 and GHC 9.4

### LINKED ISSUE #1120: Support GHC 9.4
Blocked by:
- [x] https://github.com/acid-state/safecopy/issues/88
- [ ] https://github.com/haskell/haddock/issues/1507

andreasabel:
Fix #1120.

Warning!  This PR makes `hackage-server` compile with Cabal-3.8, but does not add any Cabal-3.8 specific features.

Blocked on:
- [x] https://github.com/haskell/cabal/issues/8554

### DIFF
--- a/src/Distribution/Server/Features/BuildReports/BuildReport.hs
+++ b/src/Distribution/Server/Features/BuildReports/BuildReport.hs
@@ -1,9 +1,11 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternSynonyms #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeFamilies #-}
@@ -57,6 +59,10 @@ import Distribution.CabalSpecVersion
          ( CabalSpecVersion(CabalSpecV2_4) )
 import Distribution.Pretty
          ( Pretty(..), pretty, prettyShow )
+#if MIN_VERSION_Cabal(3,7,0)
+import Distribution.Fields.Pretty
+         ( pattern NoComment )
+#endif
 import qualified Text.PrettyPrint as Disp
 
 import Distribution.Parsec
@@ -311,7 +317,13 @@ intPair = do
 -- Pretty-printing
 
 show :: BuildReport -> String
-show = showFields (const []) . prettyFieldGrammar CabalSpecV2_4 fieldDescrs
+show = showFields noComment . prettyFieldGrammar CabalSpecV2_4 fieldDescrs
+  where
+#if MIN_VERSION_Cabal(3,7,0)
+    noComment _ = NoComment
+#else
+    noComment _ = []
+#endif
 
 -- -----------------------------------------------------------------------------
 -- Description of the fields, for parsing/printing
--- a/src/Distribution/Server/Util/CabalRevisions.hs
+++ b/src/Distribution/Server/Util/CabalRevisions.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE FlexibleContexts           #-}
 {-# LANGUAGE FunctionalDependencies     #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE PatternSynonyms            #-}
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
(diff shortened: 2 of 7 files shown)


andreasabel:
Somehow the use of `Cabal-3.8` fails with GHC 9.2 and below, with one  factor being the dependencies of `doctest-parallel`.  If in CI i fix `ghc` to the installed package, no build plan is found:
```
cabal-3.6.2.0: Could not resolve dependencies:
[__0] trying: hackage-server-0.5.1 (user goal)
[__1] rejecting: hackage-server:!test (constraint from config file, command line flag, or user target requires opposite flag selection)
[__1] trying: hackage-server:*test
[__2] trying: doctest-parallel-0.2.5 (dependency of hackage-server *test)
[__3] trying: ghc-9.2.4/installed-9.2.4 (dependency of doctest-parallel)
[__4] next goal: Cabal (dependency of hackage-server)
[__4] rejecting: Cabal-3.6.3.0/installed-3.6.3.0 (conflict: hackage-server => Cabal^>=3.8.1.0)
[__4] rejecting: Cabal-3.8.1.0 (conflict: ghc => process==1.6.13.2/installed-1.6.13.2, Cabal => process>=1.6.14.0 && <1.7)
```
I don't understand the last part: `Cabal => process>=1.6.14.0 && <1.7`.  Where does this come from?  On https://hackage.haskell.org/package/Cabal, the lower bound is `>= 1.2`.  Ok, when looking closer, at https://hackage.haskell.org/package/Cabal-3.8.1.0/dependencies, then I see this:
```
if impl(ghc>=8.2)
  process >=1.6.14.0
```
I wonder what purpose is served by this? https://hackage.haskell.org/package/process-1.6.14.0 was released 2022-02-15, but GHC 8.2 was released much earlier (in 2017).  Basically, no installed `ghc` will be compatible with such a recent version of `process`.

andreasabel:
@gbaz suggested in https://github.com/haskell/cabal/issues/8554#issuecomment-1293712012:
> for just making things work for hackage server, we can toss allow-older into the cabal.project file

This works locally (e.g. on my machine).  However, `haskell-ci` which generates our cabal-CI does not take the `cabal.project*` files into account, it actually deletes them.  Also, it does not support `allow-newer/older` settings to be passed in via the `cabal.haskell-ci` configuration file.  This somehow makes sense, as it tests the packages _as if it had been released to hackage_.  In particular it uses the `sdist` generated tarball rather than the repo code directly.  This was actually a feature I wanted, to detect files forgotten in the `hackage-server.cabal` file, see:
- https://github.com/haskell/hackage-server/pull/1061#issuecomment-1100932874

For now, I get `haskell-ci.yml` to pass by switching off the tests for ghc < 9.4.  This removes the dependency path `doctest-parallel -> ghc -> process`, so it is ok if `Cabal-3.8` wants a too recent `process`.

I don't have any clue how to fix the `nix-shell.yml` workflow.  @peterbecich, could you have a look? 
It seems to use `ghc-9.0` and seems to want to build the tests.  If you upgrade it to `ghc-9.4` it might just work.

src/Distribution/Server/Features/BuildReports/BuildReport.hs
```diff
@@ -57,6 +59,10 @@ import Distribution.CabalSpecVersion
          ( CabalSpecVersion(CabalSpecV2_4) )
 import Distribution.Pretty
          ( Pretty(..), pretty, prettyShow )
+#if MIN_VERSION_Cabal(3,7,0)
+import Distribution.Fields.Pretty
+         ( pattern NoComment )
+#endif
```
andreasabel, line 65: Alternatively, instead of `#if`, we could burn the bridges and drop Cabal-3.6 entirely.

src/Distribution/Server/Util/CabalRevisions.hs
```diff
@@ -39,6 +40,9 @@ import Distribution.Version
 import Distribution.Compiler (CompilerFlavor)
 import Distribution.FieldGrammar (prettyFieldGrammar)
 import Distribution.Fields.Pretty (PrettyField (..), showFields)
+#if MIN_VERSION_Cabal(3,7,0)
+import Distribution.Fields.Pretty (pattern NoComment)
+#endif
```
andreasabel, line 45: I want this file to remain in sync with its copy in `hackage-cli`, so the CPP would stay here.

andreasabel:
@gbaz : With CI green, maybe you want to review now?

gbaz (approved):
Looks great, thanks!

What 3.8 specific features do you think we need to consider supporting, if any?

ulysses4ever:
Multi-lib packages?

andreasabel:
> I don't have any clue how to fix the `nix-shell.yml` workflow. @peterbecich, could you have a look?
> It seems to use `ghc-9.0` and seems to want to build the tests. If you upgrade it to `ghc-9.4` it might just work.

I fixed this by adding `allow-older: Cabal:process` to the `cabal.project` file.  But anyway, there seems a 22.05 nixos available (I could be mistaken) so we could upgrade from 21.11, maybe.

andreasabel:
I'll merge this now, and open an issue for supporting Cabal-3.8 features.

Issue #1153, opened 2023-01-03
remove `LegacyPasswd` code and dependency on `crypt`


peterbecich:
`hackage-server` has a dependency on the `extra-library` `crypt`:
https://github.com/haskell/hackage-server/blob/5f5b8144ead7476f6c8af89844d84f9771106221/hackage-server.cabal#L443-L444

I believe that refers to this function: https://en.wikipedia.org/wiki/Crypt_(C)

It appears that `crypt` is deprecated:
https://www.gentoo.org/support/news-items/2021-10-18-libxcrypt-migration-stable.html

> The implementation of libcrypt.so within glibc has been deprecated
for a long time and will be removed in the near future.

> For this reason, we are following other distributions (where
this has been tested for years already) and switching to the 
external libxcrypt implementation, now also in stable installations.

It also looks like Nix is moving away from `libcrypt` to a different library `libxcrypt`: https://github.com/NixOS/nixpkgs/pull/181764

-------------------

The usage of the library in `hackage-server` is here:
https://github.com/haskell/hackage-server/blob/c3312b05eae311584e6fbbbc7201e742d87b6b9e/src/Distribution/Server/Features/LegacyPasswds/Auth.hs#L49

Can we replace `libcrypt` with `libxcrypt` , or replace this `foreign import ccall` with some other library like
- https://hackage.haskell.org/package/cryptonite
- https://hackage.haskell.org/package/crypt-sha512
- https://hackage.haskell.org/package/bcrypt-0.0.11/docs/Crypto-BCrypt.html#v:hashPassword

peterbecich:
Maybe
> libmcrypt
> Replacement for the old crypt() package and crypt(1) command, with extensions

 https://search.nixos.org/packages?channel=unstable&show=libmcrypt&from=50&size=50&sort=relevance&type=packages&query=crypt

peterbecich:
The easiest way to solve this is by deleting the "old-style crypt() auth", if possible:
https://github.com/haskell/hackage-server/blob/5f5b8144ead7476f6c8af89844d84f9771106221/src/Distribution/Server/Features/LegacyPasswds/Auth.hs#L29-L43

gbaz:
At this point I think just ripping out the legacypasswd stuff is acceptable -- the migration has been over for a long time now. PR welcome.

PR #1154, opened 2023-01-03
replace `shell.nix` with Nix Flake

peterbecich:
I think this will speed up the CI build time (https://github.com/haskell/hackage-server/issues/1078) plus other benefits

-------------

TODO 

- fix Mac building
  - https://github.com/input-output-hk/haskell.nix/issues/1825 
  - temp fix is to replace https://github.com/peterbecich/hackage-server/blob/bc488232304f63cb2ee499c32a151f601b98adb1/flake.nix#L22 with `x86_64-darwin` or `aarch64-darwin`
- https://github.com/haskell/hackage-server/issues/1153
  - see this issue with `nix flake check`
- fix DocTest
  - https://github.com/input-output-hk/haskell.nix/issues/1829
  - https://github.com/input-output-hk/haskell.nix/issues/221#issuecomment-1252180873

PR #1165, opened 2023-01-16
disable Basic Auth to remove dependency on deprecated `crypt` library

peterbecich:
Closes: https://github.com/haskell/hackage-server/issues/1153

This PR disables Basic Auth, per this comment: https://github.com/haskell/hackage-server/issues/1153#issuecomment-1370308832 Basic Auth is disabled in the simplest possible way, by returning a `UnrecognizedAuthError`. More could be done to completely delete Basic Auth. Is this sufficient for now?

`hackage-server` builds successfully without the `crypt` library.

I have barely tested this; I am able to log in with the "admin" account on Hackage running locally.

### DIFF
--- a/hackage-server.cabal
+++ b/hackage-server.cabal
@@ -441,9 +441,6 @@ library lib-server
   if flag(cabal-parsers)
     build-depends: cabal-parsers ^>= 0
 
-  if !os(darwin)
-    extra-libraries: crypt
-
 ----------------------------------------------------------------------------
 
 common exe-defaults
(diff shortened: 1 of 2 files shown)


gbaz:
I think it would be better to get rid of that file entirely and change here (https://github.com/haskell/hackage-server/blob/b4ea1c73ba85f24985259fd1df4664747ceb0159/src/Distribution/Server/Framework/Auth.hs#L103) to return a response about basic auth not being supported any longer.

peterbecich:
@gbaz , do you want `checkTokenAuth` removed, as well?
https://github.com/haskell/hackage-server/blob/b6080f390eb2896943967a006fcdc63c9ba9c390/src/Distribution/Server/Framework/Auth.hs#L103-L111

peterbecich:
@andreasabel , would you review this? Is it adequate as is? We can further remove the old auth system later IMO.

Merging this will unblock https://github.com/haskell/hackage-server/pull/1154

gbaz (approved):

PR #1168, opened 2023-01-26
Notify maintainers when dependency is out of range

ysangkok:
Tested by:
1. lowering limits as explained in `UserNotify.hs` to 1 minute
2. installing postfix, configuring it for "Localhost delivery only" and changing my Hackage user's email to `janus@myhostname.local.domain`
3. enable the notification on the preferences page, opt in to notifications
4. uploading transformers-0.5.6.2 (with patches to fix stricter cabal file validation)
5. uploading servant-0.19.1 (which doesn't accept transformers-0.6)
6. uploading transformers-0.6.0.6
7. after a minute, verified using `mail(1)` from the package `bsd-mailx` that the mail was delivered and the link works

### DIFF
--- a/src/Distribution/Server/Features/UserNotify.hs
+++ b/src/Distribution/Server/Features/UserNotify.hs
@@ -20,7 +20,7 @@ import Distribution.Version
 
 import qualified Distribution.Server.Users.Users as Users
 import Distribution.Server.Users.Group
-import Distribution.Server.Users.Types (UserId, UserInfo (..), UserName)
+import Distribution.Server.Users.Types (UserId, UserInfo (..))
 import Distribution.Server.Users.UserIdSet as UserIdSet
 
 import Distribution.Server.Packages.Types
@@ -630,8 +630,6 @@ userNotifyFeature ServerEnv{serverBaseURI, serverCron}
         idx <- queryGetPackageIndex
         revIdx <- liftIO queryReverseIndex
         let
-          userIdToName :: UserId -> Maybe UserName
-          userIdToName = fmap userName . flip Users.lookupUserId users
           genEmails :: PackageIdentifier -> IO (Map.Map (UserId, PackageId) [PackageId])
           genEmails =
             dependencyReleaseEmails (queryUserGroup . maintainersGroup) idx revIdx queryGetUserNotifyPref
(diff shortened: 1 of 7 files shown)


mitchellwrosen:
This would be an awesome feature! :tada:

src/Distribution/Server/Features/UserNotify.hs
```diff
@@ -357,12 +453,109 @@ initUserNotifyFeature env@ServerEnv{ serverStateDir, serverTemplatesDir,
...
+    Just foundPackage -> do
+      let
+        vertices :: Set.Set Vertex
+        vertices = suc revs foundPackage
+        revDepNames :: [PackageName]
+        revDepNames = mapMaybe (`lookupR` nodemap) (Set.toList vertices)
+      toNotify <- traverse maintainersToNotify revDepNames
+      pure $
+        Map.fromList
+          [ ( maintainerId
```
aaronallen8455, line 487: What happens here if multiple packages by the same maintainer are affected by the same release? Ideally, I think the maintainer should get a single notification with a list of those packages.

ysangkok: Ok, I have pushed a commit which groups by userId and packageId. Now the email looks like this:

```
The dependency transformers-0.6.0.2 has been updated.

You have requested to be notified when a dependency isn't accepted by any of your maintained packages.

These are your packages that require transformers but don't accept 0.6.0.2:

servant-0.19.1

You can adjust your notification preferences at

http://gorm.local:8080/user/admin/notify
```

gbaz:
@peterbecich any thoughts on why the nix flake test is consistently failing?

ulysses4ever:
RE nix failure: it's due to Cabal update which didn't take into account the nix setup (#1186). I tried to fix it (https://github.com/ulysses4ever/hackage-server/commit/aa00808c8cdecf32745031481ec91395e15eeeb4) but bumped into a test suite failure (https://github.com/ulysses4ever/hackage-server/actions/runs/4682410347/jobs/8296230988) and i didn't go as far as figuring what's the best way to handle it (e.g. one could probably disable the tests).

peterbecich:
Will look at the Nix error later today. Have you tried `nix flake update`? Thanks

peterbecich:
Attempting to fix: https://github.com/haskell/hackage-server/pull/1201

ysangkok:
@gbaz I have now double checked everything and I think this is ready for merging!

PR #1176, opened 2023-02-04
fix `incomplete-uni-patterns` warnings

peterbecich:
the `flake.nix` build fails on these warnings: https://github.com/haskell/hackage-server/pull/1154

No effective changes in this PR

### DIFF
--- a/exes/Main.hs
+++ b/exes/Main.hs
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -fno-warn-incomplete-uni-patterns #-}
 module Main where
 
 import qualified Distribution.Server as Server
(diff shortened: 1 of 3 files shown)


exes/Main.hs
```diff
@@ -357,15 +357,19 @@ runAction opts = do
                -> return n
       _        -> fail $ "bad port number " ++ show str
 
+    checkHostURI :: ServerConfig -> Maybe String -> Int -> IO URI
     checkHostURI defaults Nothing port = do
       let guessURI       = confHostUri defaults
-          Just authority = uriAuthority guessURI
+      case uriAuthority guessURI of
+        Nothing -> fail "No URI Authority"
```
peterbecich, line 364: this `fail` is the only change in this PR which might have some effect, is this acceptable?

andreasabel: Why not simply disable the warning also in this case?
Is there evidence that this `Just` match can actually fail?
NB: Ideally unexpected failing matches would come with source location (`HasCallStack`).

andreasabel (commented):
My general strategy would be to turn off the `incomplete-uni-patterns` warning, unless there is evidence that the match can actually fail.
(This is the conservative approach, trusting that the original code was correct.)

exes/Main.hs
peterbecich: No problem, reverted the changes and applied the pragma, like the other files

It was the easiest of the warnings to fix, and using `fail` is consistent with the rest of the file IMO

There is no evidence the `Just` match can actually fail

andreasabel (approved):

andreasabel:
Thanks @peterbecich !

PR #1179, opened 2023-02-09
Case-insensitive changelog & readme file extension check

wismill:
It is not uncommon to have files like README.TXT or NEWS.TXT.


andreasabel (approved):
Thanks!

PR #1140, opened 2022-10-16
Add lastVersion in listings (#749)

blackheaven:

PR #1184, opened 2023-02-12
Fix lastVersion update in listings

blackheaven:
Following my PR Adding lastVersion inlistings (#1140), [@ysangkok noticed they didn't update as new version get published](https://github.com/haskell/hackage-server/pull/1140#issuecomment-1409707808).

This PR attempts to fix it.

Before:

image

After:

image

Thanks again @ysangkok


blackheaven:
The failure does not seem related

ysangkok (approved):

andreasabel (approved):
Since the feature is already on `master`, I commit this fix as well.

PR #1187, opened 2023-02-27
Update link to CLC on the main page

chshersh:
The `Submitting changes for the core libraries` links to an outdated wiki page which redirects to the CLC GitHub repository.

I propose to link directly to the CLC repository

Links in question:

* https://wiki.haskell.org/Library_submissions
* https://github.com/haskell/core-libraries-committee/blob/main/PROPOSALS.md

### DIFF
--- a/datafiles/templates/index.html.st
+++ b/datafiles/templates/index.html.st
@@ -78,7 +78,7 @@ command.</li>
 <li><a href="https://wiki.haskell.org/Taking_over_a_package">Taking over a package</a></li>
 <li><a href="https://wiki.haskell.org/Abandoning_a_package">Abandoning a package</a></li>
 <li><a href="https://wiki.haskell.org/Hackage_trustees">Hackage trustees</a> and what they do</li>
-<li><a href="https://github.com/haskell/core-libraries-committee/blob/main/PROPOSALS.md">Submitting  changes</a> for the core libraries</li>
+<li><a href="https://github.com/haskell/core-libraries-committee/blob/main/PROPOSALS.md">Submitting  changes</a> for the base library</li>
 </ul>
 
 <h2 id="reporting-problems">Reporting problems</h2>
--- a/datafiles/templates/index.html.st
+++ b/datafiles/templates/index.html.st
@@ -78,7 +78,7 @@ command.</li>
 <li><a href="https://wiki.haskell.org/Taking_over_a_package">Taking over a package</a></li>
 <li><a href="https://wiki.haskell.org/Abandoning_a_package">Abandoning a package</a></li>
 <li><a href="https://wiki.haskell.org/Hackage_trustees">Hackage trustees</a> and what they do</li>
-<li><a href="https://wiki.haskell.org/Library_submissions">Submitting  changes</a> for the core libraries</li>
+<li><a href="https://github.com/haskell/core-libraries-committee/blob/main/PROPOSALS.md">Submitting  changes</a> for the core libraries</li>
 </ul>
 
 <h2 id="reporting-problems">Reporting problems</h2>


andreasabel (commented):
Thanks, @chshersh , this looks fine.

However, in general the item does not really fit in here, it is nothing specific about Hackage.  Maybe it should be removed altogether?

chshersh:
@andreasabel I asked other CLC members on [the CLC mailing list](https://groups.google.com/g/haskell-core-libraries/c/pftUMi4lh9A).

Indeed, it looks like all other links are relevant to Hackage administrative issues while CLC is not necessarily responsible for some of the Hackage parts. Although, I can see how it may seem confusing to outside collaborators when they want to contribute to a package but suddenly this package follows a completely different process, and nowhere on Hackage it's written about this process. 

Again, I lack the context for having the link to CLC on Hackage in the first place, so maybe someone remembers better.

Bodigrim:
I think it’s worth to have this link on Hackage. People do ask questions both about the role of CLC and about who is responsible for core infrastructure, so a link does not hurt.

datafiles/templates/index.html.st
```diff
@@ -78,7 +78,7 @@ command.</li>
 <li><a href="https://wiki.haskell.org/Taking_over_a_package">Taking over a package</a></li>
 <li><a href="https://wiki.haskell.org/Abandoning_a_package">Abandoning a package</a></li>
 <li><a href="https://wiki.haskell.org/Hackage_trustees">Hackage trustees</a> and what they do</li>
-<li><a href="https://wiki.haskell.org/Library_submissions">Submitting  changes</a> for the core libraries</li>
+<li><a href="https://github.com/haskell/core-libraries-committee/blob/main/PROPOSALS.md">Submitting  changes</a> for the core libraries</li>
```
hasufell, line 81:
```suggestion
<li><a href="https://github.com/haskell/core-libraries-committee/blob/main/PROPOSALS.md">Submitting  changes</a> for the base library</li>
```

Core.libraries are not under CLC purview (anymore).

andreasabel:
So if it is even just about changes to the `base` library, which is one of 10000 in the view of Hackage, why make a bullet point among basic info about submitting to Hackage?

@chshersh wrote:
> Although, I can see how it may seem confusing to outside collaborators when they want to contribute to a package but suddenly this package follows a completely different process, and nowhere on Hackage it's written about this process.

I am afraid I do not follow, because we do not explain processes how to contribute to packages published on Hackage.
So, there is no standard process we advertise, so we do not need to advertise this special process either.

Maybe the extent to which Hackage is involved in the constribution process is this:  If I want to contribute to a package, I go to _its Hackage page_, from which I get the link to the repo, and then I submit an issue or PR there.  
In particular, I don't go to the _main Hackage page_ if I want to contribute to a package.  Hackage positively does not accept any contributions to packages anyway.  It has no role in that.  Contributions happen on the respective repos.

Maybe the first step for a better entry to `base` contributors is to update the `maintainer` field of the `base` package, which still says `libraries@haskell.org`: https://hackage.haskell.org/package/base-4.18.0.0
There you could place a link to the CLC or a more specific web page.
Please facilitate that in the CLC!

I agree that it is worthwhile linking to other bodies of the Haskell infrastructure, such as the CLC and the Haskell foundation, but I think such links would be better placed in a dedicated section on bodies of the Haskell infrastructure, not among the core Hackage functionality.
I second adding a new section linking to the CLC, the Haskell foundation, and Stackage (did I forget some main body?).

Bodigrim:
@andreasabel let's not scope creep. Fixing the link is clearly an incremental improvement over status quo: pointing people to a redirection page is clearly unhelpful. We can bikeshed to death a bigger reshuffle in a separate issue.

hasufell:
> So if it is even just about changes to the base library, which is one of 10000 in the view of Hackage, why make a bullet point among basic info about submitting to Hackage?

Gosh, because it's Haskell's standard library?

andreasabel:
> Fixing the link is clearly an incremental improvement over status quo:

Clearly, so I merged this.  TODO elsewhere:
- [ ] update the maintainer field of the `base` package to the CLC link
- [ ] Consider moving the link to CLC on the Hackage entry page to a "See also" section

Bodigrim:
> * update the maintainer field of the `base` package to the CLC link

Tracked in https://gitlab.haskell.org/ghc/ghc/-/merge_requests/10259.

PR #1189, opened 2023-02-28
Implement JSON endpoint for the top downloaded packages

Kleidukos:
I tried to implement this the most direct way.

~~However I cannot start hackage-server on my machine~~
EDIT: fixed.


gbaz:
`--base-uri` doesn't set the port it connects to. you need to pass --port for that. base-uri just sets the base uri in generated html in templates. the confusing issue (pr welcome) is it just doesn't print out what port its trying to connect to which is already in use, which is typically going to be 8080 by default i think, but may be used by other things in your system since its common.

Kleidukos:
Now that the above matter is fixed, I tested the endpoint locally and it works. :)

gbaz:
@Kleidukos I think this is in shape. Do you want it merged? If so it needs to be unmarked from draft status.

Kleidukos:
Yep, ready for your approval :)

PR #1194, opened 2023-03-15
use `haskell/actions` GitHub Action

peterbecich:
https://github.com/haskell/actions/tree/main/setup


andreasabel (commented):
@peterbecich What are the motivations to change back from an auto-generated CI to a manual one?
Note that it is not long ago that I switched to Haskell-CI, and I had reasons.  Haskell-CI is more thorough: It includes `cabal haddock`, `cabal check`, and most importantly, it builds the project from the generated tarball, not directly from the repo contents.  This ensures that the package will build also if shipped (e.g. to hackage), not just when living in its "home environment".

.github/workflows/cabal.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Setup Haskell
+        uses: haskell/actions/setup@v2
+        with:
+          ghc-version: ${{ matrix.ghc }}
+          cabal-version: ${{ matrix.cabal }}
+      - name: Install dependencies
+        run: sudo apt install -y libbrotli-dev libgd-dev
+      - name: Build
+        run: cabal build all --enable-tests
+      - name: Test
```
andreasabel, line 32: No caching of the Haskell dependencies?

PR #1200, opened 2023-04-22
Undoes stray x from 5f5b8144ea

ysangkok:
Removes a character that was (errorneusly, I think) inserted by @peterbecich in commit 5f5b8144ea in #1008


peterbecich:
Thanks, yes, it was an error

Issue #1199, opened 2023-04-14
UI: Sorting table on tags view adds hard-to-read background colour


fendor:
![image](https://user-images.githubusercontent.com/8463814/232042618-37b3b611-a869-4a60-b8e4-5bc50be8ca96.png)

It looks correct without any sort applied.

Similar issue in the footer for the page numbers:
![image](https://user-images.githubusercontent.com/8463814/232042803-16fc87ac-d0d9-4a1d-9653-9208c4fd5683.png)
but there is actually more than one page, which you only see if you hover accidentally over that location:
![image](https://user-images.githubusercontent.com/8463814/232042924-0d090009-6944-4e35-9239-61c95d23116e.png)

Likely, this issue is only present in dark mode.

ffaf1:
Can reproduce on `Mozilla Firefox 102.9.0esr`. Note that this happens only on "tag" pages, not on normal searches:

![image](https://user-images.githubusercontent.com/87714215/232070948-36e60be6-89d3-43b8-956b-87e8c9e662e1.png)

gbaz:
@peterbecich

peterbecich:
I believe this fixes the sorted column background color: https://github.com/haskell/hackage-server/pull/1207
I have not looked at the other issues yet. Let's leave the issue open and solve those later. Please review.

peterbecich:
More improvements: https://github.com/haskell/hackage-server/pull/1281

PR #1207, opened 2023-05-21
Sorted table column background color

peterbecich:
https://github.com/haskell/hackage-server/issues/1199

The issue was that `datatables.min.css` took precedence over `hackage.css`. I have reversed this. I have not yet checked for other problems that may be caused by this change.

Screenshot 2023-05-21 at 14-48-11 All packages by name Hackage

TODO fix the other issues mentioned
In a different PR

### DIFF
--- a/datafiles/templates/Html/candidate-index.html.st
+++ b/datafiles/templates/Html/candidate-index.html.st
@@ -2,9 +2,9 @@
 <html>
 
     <head>
-        $hackageCssTheme()$
         <script src="/static/jquery.min.js"></script>
         <link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/v/dt/dt-1.11.4/sp-1.4.0/datatables.min.css"/>
+        $hackageCssTheme()$
 
         <script type="text/javascript" src="https://cdn.datatables.net/v/dt/dt-1.11.4/sp-1.4.0/datatables.min.js"></script>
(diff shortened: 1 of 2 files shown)


peterbecich:
At a glance, the table and candidate interfaces look okay

Screenshot 2023-05-26 at 23-24-12 Browse and search packages Hackage
Screenshot 2023-05-26 at 23-24-30 Package candidates Hackage

ffaf1 (approved):
Good!

ysangkok:
I don't understand what this is fixing, since the linked issue is solely about tags, which are implemented differently. The tags listing is created on the server, the browse page uses a JSON api. I saw your comment in the issue thread, saying that this is only supposed to fix one issue, but if I understand the linked issue correctly, all of the issues are on the tags page. So any test should involve the tags page. At least in one of the screenshots, I see 'Advanced options', which are not available on the tags page. So I don't understand how that screenshot can be relevant.

peterbecich:
You're right, thanks, the screenshot was the wrong page. Here is the tags page with the fix:
Screenshot 2023-05-27 at 11-50-03 All packages by name Hackage

ysangkok (approved):
Ok, looks good! I wonder if using specificity would be an alternative, since more specific CSS rules override less specific rules, even if they come first. But I am fine with switching the order, it's unlikely to break, I think.

Issue #1213, opened 2023-05-29
Cannot revise a package without putting in all upper bounds


Bodigrim:
I'm trying to revise https://hackage.haskell.org/package/MuCheck-0.3.0.0/MuCheck.cabal/edit with `haskell-src-exts <1.18`. However Hackage complains: 

> Errors
> These packages miss upper bounds: - directory - hashable - hint - hspec - mtl - random - syb - temporary - time Please add them, using `cabal gen-bounds` for suggestions. For more information see: https://pvp.haskell.org/ 

This should not be an error, at most a warning, however I'd suggest ignoring it automatically in the interface for revisions. 

A Hackage trustee cannot reasonably put in all upper bounds for a decade-old package, this is very annoying.

phadej:
The culprit is in https://github.com/haskell/hackage-server/blob/dc5dda308e015118e0eb1af3f4bfd29a24e17d63/src/Distribution/Server/Util/CabalRevisions.hs#L179-L185

There we compare that no new `cabal check` diagnostic are introduced. But here we swap. If each package were separate warning, things would just worked.

I hope this is resolved soon, in however hacky way necessary, as this indeed makes revision making essentially impossible.

ffaf1:
An option is to copy — or even better, abstract — what `Unpack.hs` does:

https://github.com/haskell/hackage-server/blob/dc5dda308e015118e0eb1af3f4bfd29a24e17d63/src/Distribution/Server/Packages/Unpack.hs#L297-L307

Missing upper bounds for non critical packages (`base`, `Cabal`, …) are `PackageDistSuspiciousWarn`, so they would not throw an error.

phadej:
@ffaf1 No. Filtering too much warnings will open doors for mistakes (or abuse).

gbaz:
pr at #1214

gbaz:
will redeploy this weekend

gbaz:
deploy completed. i just successfully performed a revision that was unblocked by the pr, so tested and working.

phadej:
Thanks!

PR #1214, opened 2023-06-01
relax upper bounds revision check

gbaz:
resolves https://github.com/haskell/hackage-server/issues/1213


src/Distribution/Server/Util/CabalRevisions.hs
```diff
@@ -179,7 +179,12 @@ checkCabalFileRevision checkXRevision old new = do
     checkPackageChecks :: Check GenericPackageDescription
     checkPackageChecks pkg pkg' =
       let checks  = checkPackage pkg  Nothing
-          checks' = checkPackage pkg' Nothing
+          checks' = filter notUpperBounds $ checkPackage pkg' Nothing
+          -- if multiple upper bounds are missing, then the simple set subtraction might detect a change to
+          -- just one, and fail. Ideally we'd peform a set subtraction directly on just the missing bounds
```
Bodigrim, line 184:
```suggestion
          -- just one, and fail. Ideally we'd perform a set subtraction directly on just the missing bounds
```

ffaf1 (approved):

PR #1220, opened 2023-06-19
enable `nix flake check` in GitHub Action

peterbecich:
Thanks @srid


peterbecich:
This PR shows there is an issue with the `nix develop` shell on Mac. It is not a bug in `haskell-flake`, it is misconfiguration in `flake.nix`. More info https://github.com/srid/haskell-flake/discussions/172#discussioncomment-6219087

peterbecich:
https://github.com/haskell/hackage-server/blob/cb3e0b64d5a3c42aa68a1477012d80acdf37edde/.github/workflows/nix-flake.yml#L33-L35

PR #1217, opened 2023-06-06
haiku: add as known platform

jessicah:

PR #1221, opened 2023-06-20
Recognise `OtherOS "haiku"` as a valid identifier.

jessicah:
`OSHaiku` will be available in cabal-3.12, and a PR for that change is in #1217. Until then, allow `OtherOS "haiku"` until cabal can be updated.

### DIFF
--- a/src/Distribution/Server/Framework/Instances.hs
+++ b/src/Distribution/Server/Framework/Instances.hs
@@ -119,6 +119,8 @@ instance SafeCopy VersionRange where
 instance SafeCopy OS where
     errorTypeName _ = "OS"
 
+    putCopy (OtherOS "haiku") = contain $ putWord8 18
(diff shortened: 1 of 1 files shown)


jessicah:
Corresponding PR to cabal: https://github.com/haskell/cabal/pull/9048

jessicah:
@andreasabel ah okay, was hoping to get done sooner, so that dependencies for GHC could be uploaded, as they're still waiting as PRs for a few months now. That was part of the reason for the `OtherOS "haiku"` patch, so that hackage didn't need to wait for a new major version to cabal.

andreasabel:
@jessicah : With eabf54a2e1ed56fa8cee13bea6d71680a4535267 your PR is now on `master`.  
However, it might take some time to bring it live; `hackage.haskell.org` is only deployed every couple of months, only @gbaz knows when will be next.

Issue #1222, opened 2023-06-21
Public package repositories best practice


glyn:
I think it would be best practice to make any repository associated with a Hackage package public. This would help users see current issues (e.g. to avoid bothering package maintainers with dups), explore version history, etc.

For example, the [sliceofpy](https://hackage.haskell.org/package/sliceofpy) package points to a private repository for its home page, issue tracker, and source repo. (Please note I have no complaints with this package - it was just an example I came across.)

I wonder if it would therefore be helpful to document such a best practice in the Hackage docs?

Please note I am not suggesting any kind of policy change; private repositories or tarballs with no repository would still be totally acceptable. But a docs change would gently encourage maintainers to open up their repositories.

I'll submit a PR with some suggested text.

gbaz:
should be resolved by above linked pr.

PR #1223, opened 2023-06-21
Prefer public repositories

glyn:
Fixes https://github.com/haskell/hackage-server/issues/1222

### DIFF
--- a/datafiles/templates/upload.html.st
+++ b/datafiles/templates/upload.html.st
@@ -97,9 +97,9 @@ license files of any 3rd party code that you include.</p>
 
 It is preferable if any web pages linked from the following fields are publicly accessible:
 <ul>
-<li> Bug report URL
-<li> Homepage URL
-<li> Source repositories
+<li>Bug report URL</li>
+<li>Homepage URL</li>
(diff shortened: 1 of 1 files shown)


datafiles/templates/upload.html.st
```diff
@@ -93,6 +93,22 @@ service.</p>
 material that you upload. For example, check that your tarball includes the
 license files of any 3rd party code that you include.</p>
 
+<h3>Public repositories</h3>
+
+It is preferable if any web pages linked from the following fields are publicly accessible:
+<ul>
+<li> Bug report URL
```
ffaf1, line 100: I suspect: missing `</li>`.

glyn: Fixed, thanks. (Too used to markdown these days.)

gbaz:
looks good, ty!

glyn:
Thanks for merging, @gbaz. https://hackage.haskell.org/upload hasn't been updated, so does some manual process need to be run?

gbaz:
Yeah, docs are updated on redeploys, which are irregular, unless there's a burning need :-)

PR #1224, opened 2023-06-22
Update max cabal-version check to 3.6

ffaf1:
Also add a more informative comment and abstract check/error message.

### DIFF
--- a/src/Distribution/Server/Packages/Unpack.hs
+++ b/src/Distribution/Server/Packages/Unpack.hs
@@ -216,9 +216,9 @@ specVersionChecks specVerOk specVer = do
   when (specVer < CabalSpecV1_2) $
     throwError "'cabal-version' must be at least 1.2"
 
-  -- Safeguard; should already be caught by parser
-  unless (specVer <= CabalSpecV3_0) $
-    throwError "'cabal-version' must be at most 3.0"
+  -- To keep people from uploading packages most users cannot use.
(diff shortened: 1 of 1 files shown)


Mikolaj (approved):
LGTM

Mikolaj:
Though

```
src/Distribution/Server/Packages/Unpack.hs:222:54: error:
    • Variable not in scope:
        showCabalSpecVersion :: CabalSpecVersion -> [()]
    • Perhaps you want to add ‘showCabalSpecVersion’ to the import list
      in the import of ‘Distribution.CabalSpecVersion’
      (src/Distribution/Server/Packages/Unpack.hs:(23,1)-(24,33)).
```

ulysses4ever:
Why not 3.8? i don't think it's a good reason to hold back package authors. If a user doesn't have a recent enough cabal, they should have a choice: either update their cabal or not use the package (version). It's absolutely nonsensical to me to forbid package authors to upload such packages — it should be a user's choice.

ffaf1:
Which `cabal` do I need to use `cabal-version: 3.8`? A line has to be drawn somewhere

ulysses4ever:
The principle of least surprise would suggest to use the same version as the Cabal library. Every package page show you which Cabal library version is used by Hackage (see example below). Naturally, people would expect that the same version is the maximum that you can put in your cabal file...
Screenshot_20230622-181507

PR #1233, opened 2023-07-08
Fix package list when user has multiple packages

ysangkok:
This fixes an issue @AndreasAbel encountered where you'd only get notified for one of the packages that you maintain, instead of all of them, as you'd expect.

Greetings from MuniHac!


src/Distribution/Server/Features/UserNotify.hs
```diff
@@ -473,7 +473,7 @@ dependencyReleaseEmails userSetIdForPackage index (ReverseIndex revs nodemap dep
         revDepNames = mapMaybe (`lookupR` nodemap) (Set.toList vertices)
       toNotify <- traverse maintainersToNotify revDepNames
       pure $
-        Map.fromList
+        Map.fromListWith (++)
```
andreasabel, line 476: Yeah, `Map.fromList` is harmful! :-)

andreasabel (approved):
Thanks!

Issue #1237, opened 2023-08-05
Nix-workflow broken


andreasabel:
Possibly due to the addition of dependency `attoparsec-aeson`, the nix workflow is broken:
https://github.com/haskell/hackage-server/actions/runs/5769695862/job/15642001016#step:5:94
```
building '/nix/store/bia6q1zrmnqx405bx3q54gh0sr7vjcrp-cabal2nix-hackage-server.drv'...
error:
 ...
       error: function 'anonymous lambda' called without required argument 'attoparsec-aeson'

       at /nix/store/p5xpdv1kr2kdwp1cga6ss7cxmfzdb43v-cabal2nix-hackage-server/default.nix:1:1:

            1| { mkDerivation, acid-state, aeson, array, async, attoparsec
             | ^
            2| , attoparsec-aeson, attoparsec-iso[86](https://github.com/haskell/hackage-server/actions/runs/5769695862/job/15642001016#step:5:87)01, base, base16-bytestring
Error: Process completed with exit code 1.
```
ATTN: @peterbecich

peterbecich:
Need to get `attoparsec-aeson` into NixPkgs: https://github.com/NixOS/nixpkgs/issues/247406
I will also try to provide this packages is a Flake Input.

peterbecich:
Facing an issue with overriding `attoparsec-aeson`, asked for help: https://github.com/srid/haskell-flake/discussions/188

peterbecich:
Fixed for Linux: https://github.com/haskell/hackage-server/pull/1241
But need to fix Mac separately

peterbecich:
Mac Flake fixed here: https://github.com/haskell/hackage-server/pull/1244

peterbecich:
Closing this issue. `nix develop` on Mac TODO: https://github.com/haskell/hackage-server/issues/1247

PR #1241, opened 2023-08-13
fix `flake.nix` for Linux

peterbecich:
All of these should work now:

- `nix develop`
- `nix build`
- `nix flake check`

https://github.com/haskell/hackage-server/issues/1237

The only risky change is this PR is lowering the lower bound on `aeson` from `2.2.*` to `2.1.*`.  
It's necessary because Nix is still on `aeson` 2.1. This is true for both `nixpkgs-unstable` and the newer branch `haskell-updates`: https://github.com/NixOS/nixpkgs/tree/haskell-updates

Is this okay? 

---------------

The Flake is still broken on Mac. This looks more complicated to fix. I'll try to fix that in a subsequent PR.

### DIFF
--- a/hackage-server.cabal
+++ b/hackage-server.cabal
@@ -135,7 +135,7 @@ common defaults
     , scientific
   -- other dependencies shared by most components
   build-depends:
-    , aeson                  >= 2.1.0.0
+    , aeson                  >= 2.1.0.0 && < 2.3
     , Cabal                  >= 3.10.1.0 && < 3.12
     , Cabal-syntax           >= 3.10.1.0 && < 3.12
         -- Cabal-syntax needs to be bound to constrain hackage-security
@@ -553,7 +553,7 @@ test-suite HighLevelTest
   build-depends:
     -- version constraints inherited from lib-server
     , HTTP
-    , attoparsec-aeson >= 2.1.0.0
+    , attoparsec-aeson >= 2.1.0.0 && < 2.3
     , base64-bytestring
     , random
       -- component-specific dependencies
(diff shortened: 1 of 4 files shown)


peterbecich:
Succeeds:
```
cabal build all --constraint 'aeson < 2.2' --constraint 'attoparsec-aeson < 2.2' --enable-tests
```

@andreasabel please review my change to `aeson` and `attoparsec-aeson` lower bounds: https://github.com/haskell/hackage-server/pull/1241/files#diff-80d8280eac7e7395c335fe96148a837a1e7bcf1f8e3511fb84465551ed80cf47

hackage-server.cabal
```diff
@@ -135,7 +135,7 @@ common defaults
     , scientific
   -- other dependencies shared by most components
   build-depends:
-    , aeson                 ^>= 2.2.0.0
+    , aeson                  >= 2.1.0.0
```
andreasabel, line 138: All dependencies have upper bounds here, so if you switch from `^>=` to `>=` you need an upper bound, like `>= 2.1.0.0 && < 2.3`.
Alternatively, you can list just both tested versions: `^>= 2.1.0.0 || ^>= 2.2.0.0`

hackage-server.cabal
```diff
@@ -553,7 +553,7 @@ test-suite HighLevelTest
   build-depends:
     -- version constraints inherited from lib-server
     , HTTP
-    , attoparsec-aeson ^>= 2.2.0.0
+    , attoparsec-aeson >= 2.1.0.0
```
andreasabel, line 556: ditto

andreasabel (commented):
Thanks @peterbecich !

The relaxation of `aeson` LGTM, but an upper bound should be kept (see my comments).

andreasabel (commented):
Thanks!

PR #1243, opened 2023-08-17
Provide both plain text + HTML versions for notification emails

### LINKED ISSUE #1242: Dependency notifications has incorrect link
An email I got yesterday:

> The dependency bytestring-0.12.0.2 has been updated.
>
> You have requested to be notified when a dependency isn't accepted by any of your maintained packages.
>
> These are your packages that require bytestring but don't accept [0.12.0.2](http://0.12.0.2/):
>
> github-rest-1.1.4

The `0.12.0.2` link gets linked to http://0.12.0.2, which seems... wrong. Instead of linking `0.12.0.2`, it'd be great if `bytestring-0.12.0.2` + `github-rest-1.1.4` could be linked to https://hackage.haskell.org/package/bytestring-0.12.0.2 + https://hackage.haskell.org/package/github-rest-1.1.4

brandonchinn178:
Fixes #1242 

Apologies in advance for the huge changeset. Probably easiest to review commit-by-commit.

Everything up to the "Add markup to emails" commit is what's minimally necessary to send both plain text + HTML. The commits after "Add markup to emails" include a ton of refactoring in order to test this. Happy to break this out into a separate PR if desired.

Key areas of note:
* "Add markup to emails" (e0c89e7c): This adds markup like linking things, making bulleted lists, etc. Would be good to audit this. The HTML can be inspected in the [golden files](https://github.com/brandonchinn178/hackage-server/tree/notify-email/tests/golden/ReverseDependenciesTest).
* "Simplify describeDependencyUpdate" (https://github.com/haskell/hackage-server/pull/1243/commits/d0d539ef0dc9bc866f30c2740c42c49089e389ee): This commit slightly changes the template sent out for dependency updates. It makes the rendering logic more linear, at the cost of a bit more verbosity.
* "Improve email subject for dep update emails" (https://github.com/haskell/hackage-server/pull/1243/commits/6a46f306f609eed4d66dfa3c5e5dfe5889fad1c3): Instead of a generic "Maintainer Notifications" subject line, dependency update emails will have the dependency name in the subject line.
    * Related, why not batch all the emails, or batch none of them? Now that we're rendering HTML, we can use `<hr>` to render a horizontal line between notifications (plain text could just render a line of hyphens). But it also doesn't seem likely for multiple notifications to come at once; maybe batching is premature optimization + sending one email per notification is fine?

Happy to set up a call if you want to pair review. I'm on PST, my email can be found in my profile

### DIFF
--- a/src/Distribution/Server/Features/UserNotify.hs
+++ b/src/Distribution/Server/Features/UserNotify.hs
@@ -702,10 +702,8 @@ userNotifyFeature ServerEnv{serverBaseURI, serverCron}
           genEmails :: PackageIdentifier -> IO (Map.Map (UserId, PackageId) [PackageId])
           genEmails =
             dependencyReleaseEmails (queryUserGroup . maintainersGroup) idx revIdx queryGetUserNotifyPref
-        dependencyEmailMaps <- traverse (genEmails . pkgInfoToPkgId) revisionsAndUploads
+        dependencyEmailMap <- Map.unionsWith (++) <$> traverse (genEmails . pkgInfoToPkgId) revisionsAndUploads
         let
-          dependencyEmailMap :: Map.Map (UserId, PackageId) [PackageId]
-          dependencyEmailMap = Map.unionsWith (++) dependencyEmailMaps
           emailText :: MonadIO m => (UserId, PackageId) -> [PackageId] -> m [String]
           emailText (uId, dep) revDeps = do
             mPrefs <- queryGetUserNotifyPref uId
--- a/src/Distribution/Server/Features/UserNotify.hs
+++ b/src/Distribution/Server/Features/UserNotify.hs
@@ -699,8 +699,7 @@ userNotifyFeature serverEnv@ServerEnv{serverBaseURI, serverCron}
         let docReportEmails = foldMap describeDocReport <$> docReportNotifications
 
         tagProposals <- collectTagProposals
-        tagProposalNotifications <- foldM (genTagProposalList notifyPrefs) Map.empty tagProposals
-        let tagProposalEmails = foldMap describeTagProposal <$> tagProposalNotifications
+        tagProposalNotifications <- concatMapM (genTagProposalList notifyPrefs) tagProposals
 
         idx <- queryGetPackageIndex
         revIdx <- liftIO queryReverseIndex
@@ -711,13 +710,13 @@ userNotifyFeature serverEnv@ServerEnv{serverBaseURI, serverCron}
           getNotificationEmails serverEnv userDetailsFeature users
             ( foldr1 (Map.unionWith (<>))
               [ docReportEmails
-              , tagProposalEmails
               ]
             , dependencyEmails
             ) $
             concat
               [ revisionUploadNotifications
               , groupActionNotifications
+              , tagProposalNotifications
               ]
         mapM_ sendNotifyEmailAndDelay emails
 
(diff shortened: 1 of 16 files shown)


src/Distribution/Server/Features/UserNotify.hs
```diff
@@ -817,18 +832,6 @@ userNotifyFeature serverEnv@ServerEnv{serverBaseURI, serverCron}
...
-      | pkgNumRevisions pkg <= 1 =
-          EmailContentParagraph $
-            "Package upload, " <> renderPkgLink (pkgInfoId pkg) <> ", by " <>
-            formatTimeUser users (pkgLatestUploadTime pkg) (pkgLatestUploadUser pkg)
-      | otherwise =
-          EmailContentParagraph ("Package metadata revision(s), " <> renderPkgLink (pkgInfoId pkg) <> ":")
-          <> EmailContentList (map (uncurry (formatTimeUser users) . snd) recentRevs)
-        where
-           revs = reverse $ Vec.toList (pkgMetadataRevisions pkg)
-           recentRevs = filter ((\x -> x > earlier && x <= now) . fst . snd) revs
```
brandonchinn178, line 842: Previously, the logic was broken up as:
1. `collectRevisionsAndUploads`: Filter packages if they have recent revisions
2. `genRevUploadList`: Filter packages if user should be notified
3. `describeRevision`: Find the recent revisions + render them into message

Now, finding the recent revisions is moved into `genRevUploadList` so that `renderNotifyNewRevision` can just be responsible for rendering into an email.

src/Distribution/Server/Features/UserNotify.hs
```diff
@@ -797,18 +794,36 @@ userNotifyFeature serverEnv@ServerEnv{serverBaseURI, serverCron}
...
-            addNotification uid m = if not (notifyOptOut npref) && notifyMaintainerGroup npref
-                                      then Map.insertWith (++) uid [ga] m
-                                      else m
-                where npref = fromMaybe defaultNotifyPrefs (Map.lookup uid notifyPrefs)
-        in case gdesc of
-           (MaintainerGroup pkg) -> do
-              maintainers <- queryUserGroup $ maintainersGroup (mkPackageName $ BS.unpack pkg)
-              return $ foldr addNotification mp (toList (delete actor maintainers))
-           _ -> return mp
+    genGroupUploadList notifyPrefs groupAction =
```
brandonchinn178, line 789: All of these functions generating `Notification` values should probably be broken out + tested like `getUserNotificationsOnRelease`. But this PR is big enough already 🙃

src/Distribution/Server/Features/UserNotify.hs
```diff
@@ -899,36 +898,103 @@ userNotifyFeature ServerEnv{serverBaseURI, serverCron}
...
-            <> emailContentUrl
-                serverBaseURI
-                  { uriPath =
-                      concatMap ("/" <>)
-                        [ "user"
-                        , display $ Users.userIdToName users uid
-                        , "notify"
-                        ]
-                  }
+      -- delay sending out emails, because ???
```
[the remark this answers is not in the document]
ysangkok: It's something @gbaz requested, I think it's because we are afraid that we'd be labelled as spammers if we somehow accidentally sent out too many e-mails.

src/Distribution/Server/Util/Email.hs
```diff
@@ -0,0 +1,148 @@
...
+emailContentIntercalate x = mconcat . intersperse x
+
+emailContentUrl :: URI -> EmailContent
+emailContentUrl uri = EmailContentLink (uriToText uri) uri
+
+fromEmailContent :: EmailContent -> Alternatives
+fromEmailContent emailContent =
+  [ Part
+      { partType = contentType <> "; charset=utf-8"
+      , partEncoding = None
```
ysangkok, line 73: I know that we also used `None` in the old version, but I am wondering whether it wouldn't be better to use `QuotedPrintableText` as mime-mail does itself: [Network.Mail.Mime.htmlPart](https://hackage.haskell.org/package/mime-mail-0.5.1/docs/src/Network.Mail.Mime.html#htmlPart). I am not sure we can actually generate any non-ASCII characters, so maybe it is irrelevant. But QuotedPrintableText does seem safer.

brandonchinn178: Oh!! I should just use `htmlPart` and `plainPart` 😄

ysangkok:
It looks good to me, have you tested it with a local deployment? You can use a local Postfix installation.

brandonchinn178:
@ysangkok Sorry, I can't seem to get postfix working on my Mac. Would it be possible for you to test it locally?

ysangkok:
Tested locally with packages transformers-0.5/0.6 and servant-0.19/0.20 and works as expected. Patched the email generation as follows:

```diff
diff --git a/src/Distribution/Server/Features/UserNotify.hs b/src/Distribution/Server/Features/UserNotify.hs
index fb11fa5e..032e2b75 100644
--- a/src/Distribution/Server/Features/UserNotify.hs
+++ b/src/Distribution/Server/Features/UserNotify.hs
@@ -680,7 +680,7 @@ userNotifyFeature serverEnv@ServerEnv{serverCron}
     setupNotifyCronJob =
       addCronJob serverCron CronJob {
         cronJobName      = "send notifications",
-        cronJobFrequency = TestJobFrequency (60*60*2), -- 2hr (for testing you can decrease this)
+        cronJobFrequency = TestJobFrequency 10, -- 2hr (for testing you can decrease this)
         cronJobOneShot   = False,
         cronJobAction    = notifyCronAction
       }
@@ -688,9 +688,7 @@ userNotifyFeature serverEnv@ServerEnv{serverCron}
     notifyCronAction = do
         (notifyPrefs, lastNotifyTime) <- unNotifyData <$> queryState notifyState GetNotifyData
         now <- getCurrentTime
-        let trimLastTime = if diffUTCTime now lastNotifyTime > (60*60*6) -- cap at 6hr
-                             then addUTCTime (negate $ (60*60*6)) now
-                             else lastNotifyTime -- for testing you can increase this
+        let trimLastTime = addUTCTime (negate $ (60*60*6)) now
         users <- queryGetUserDb
 
         revisionsAndUploads <- collectRevisionsAndUploads trimLastTime now
```

brandonchinn178:
Thank you!

PR #1219, opened 2023-06-19
enable  Mac build in GitHub Action

peterbecich:
https://github.com/haskell/hackage-server/issues/1193

`nix flake check` restricted to current system in latest Nix version, so there should be no error "Mac required but I am a Linux", or vice versa

Issue #1193, opened 2023-03-14
MacOS GitHub Action minutes


peterbecich:
MacOS GitHub Actions cost more: https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers

What is our allocation for GitHub Actions? Can we afford to build this `flake.nix` for Mac?

The benefit of enabling the Mac build is that Mac artifacts will be pushed to Cachix and save time for these developers:
https://github.com/haskell/hackage-server/blob/7d357fc874aaced1e7d2aa4061bd333560c12c85/.github/workflows/nix-flake.yml#L14-L16

These initial GitHub Actions builds took 25 minutes: https://github.com/haskell/hackage-server/actions/workflows/nix-flake.yml I think the cache was not yet populated. I expect future builds will be faster. Nonetheless, the 10x multiplier on the Mac build is quite a bit.

andreasabel:
Are we paying for actions @gbaz ?  (I was assuming this was a free service for us.)

gbaz:
We're not paying for actions, but we may have limits to our shared free plan. I confess I do not know the details and would have to dig in to figure that out.

That said, I'm a macos developer but I don't use nix. We have very few developers, and I think fewer still use nix, of which I don't know how many would use cachix. I wouldn't necessarily think that it would be important in general to have this.

gbaz:
As a followup, having dug in, I think we have the action minutes to spare for now -- so while I personally wouldn't get any use out it at the moment, if there are people who would, I think it's fine, and if we start to run into limits I didn't expect, we can reconsider.

andreasabel:
> having dug in, I think we have the action minutes to spare for now

Are these minutes accounted for all of @haskell or just this repo?

srid:
> The benefit of enabling the Mac build is that Mac artifacts will be pushed to Cachix and save time for these developers

Only for Intel macs, though; see https://github.com/actions/runner-images/issues/2187

PR #1244, opened 2023-08-21
update `flake.nix` for recent `haskell-flake`

peterbecich:
- The `flake.nix` is updated to be compatible with the latest https://github.com/srid/haskell-flake

- Small tweak to `cabal.project` to fix a parsing error in the Nix Flake. No effect on `cabal`.

- `nix flake update`

- Updates to the Nix section of the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,25 +7,81 @@ This is the `hackage-server` code. This is what powers <http://hackage.haskell.o
 
 ## Installing dependencies
 
-`hackage-server` depends on `libgd` and `zlib`. You'll also need `libbrotli-dev` for enabling tests.
+`hackage-server` depends on `libgd`, `zlib`, and other system libraries. You'll also need `libbrotli-dev` for enabling tests.
 
-### [`nix develop`](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-develop.html)
+You can use the Nix package manager to provide these dependencies, or install them manually.
 
-If you have the [Nix package manager](https://nixos.org/) installed, the easiest way to run `hackage-server` is by using `nix develop`. It should be unnecessary to install any dependencies manually. In this repository:
+### Using the [Nix package manager](https://nixos.org/) and provided [Nix Flake](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake.html)
+
+If you have the Nix package manager installed, you can build and run `hackage-server` without manually installing any dependencies.
+
+This uses `flake.nix`, implemented with [`srid/haskell-flake`](https://github.com/srid/haskell-flake).
+
+There are at least three ways to use this `flake.nix`. Clone this repository, enter the repository directory, then choose one of these options:
+
+#### [`nix develop`](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-develop.html)
 
     nix develop
 
     (in develop shell)
-    $ cabal v2-run -- hackage-server init
+    $ cabal v2-run -- hackage-server init --static-dir=datafiles
 
-    $ cabal v2-run -- hackage-server run --static-dir=datafiles/ --base-uri=http://127.0.0.1:8080
+    $ cabal v2-run -- hackage-server run --static-dir=datafiles --base-uri=http://127.0.0.1:8080
     hackage-server: Ready! Point your browser at http://127.0.0.1:8080
 
-`flake.nix` is provided; it uses [`srid/haskell-flake`](https://github.com/srid/haskell-flake).
+Note the `init` command will create a new folder `state` in your working directory.
 
 If you have [direnv](https://direnv.net/), `direnv allow` will load this `nix develop` shell automatically.
 
-`nix build` will build a `hackage-server` executable in `result/`. The Hackage dependencies are provided by the inputs specified in `flake.nix`. Because some of these inputs are unpublished commits on GitHub, this build should not be considered authoritative.
+#### [`nix build`](https://nixos.org/manual/nix/stable/command-ref/nix-build.html)
+
+    nix build
+
+This will produce a `hackage-server` executable in `result/`.
+
+For this executable, Hackage dependencies are not pulled from Hackage directly like usual. Hackage dependencies are provided by the [Nixpkgs](https://search.nixos.org/packages) [`haskell-updates`](https://github.com/NixOS/nixpkgs/tree/haskell-updates) branch, and a few [overrides in `flake.nix`](https://zero-to-flakes.com/haskell-flake/dependency#using-hackage-versions).
+
+#### [`nix run`](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-run)
+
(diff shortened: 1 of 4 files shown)


peterbecich:
@andreasabel @gbaz , I have re-written the Readme section about Nix. Your reviews are appreciated:
https://github.com/haskell/hackage-server/blob/06f831ef7a701967f22128087b3817cdff185cbc/README.md#using-the-nix-package-manager-and-provided-nix-flake
All of those commands have been tested.

I understand there is a very small number of Nix developers working on this project: https://github.com/haskell/hackage-server/issues/1193#issuecomment-1501262936 The Nix section is verbose. That being said, having this documentation could potentially help gain more users of Nix Flakes. IMO Flakes are a very worthwhile development and I'd like to help draw attention to them. Ultimately I think Flakes will help Haskell greatly.

README.md
```diff
@@ -7,25 +7,81 @@ This is the `hackage-server` code. This is what powers <http://hackage.haskell.o
...
-If you have the [Nix package manager](https://nixos.org/) installed, the easiest way to run `hackage-server` is by using `nix develop`. It should be unnecessary to install any dependencies manually. In this repository:
+### Using the [Nix package manager](https://nixos.org/) and provided [Nix Flake](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-flake.html)
+
+If you have the Nix package manager installed, you can build and run `hackage-server` without manually installing any dependencies.
+
+This uses `flake.nix`, implemented with [`srid/haskell-flake`](https://github.com/srid/haskell-flake).
+
+There are at least three ways to use this `flake.nix`. Clone this repository, enter the repository directory, then choose one of these options:
+
+#### [`nix develop`](https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-develop.html)
```
andreasabel, line 22: I tried `nix develop` on my Mac Monterey and got this:
```
$ nix develop  --extra-experimental-features nix-command --extra-experimental-features flakes
do you want to allow configuration setting 'allow-import-from-derivation' to be set to 'true' (y/N)? y
do you want to permanently mark this value as trusted (y/N)? n
do you want to allow configuration setting 'extra-substituters' to be set to 'https://hackage-server.cachix.org/' (y/N)? y
do you want to permanently mark this value as trusted (y/N)? n
do you want to allow configuration setting 'extra-trusted-public-keys' to be set to 'hackage-server.cachix.org-1:iw0iRh6+gsFIrxROFaAt5gKNgIHejKjIfyRdbpPYevY=' (y/N)? y
do you want to permanently mark this value as trusted (y/N)? n
warning: ignoring untrusted substituter 'https://hackage-server.cachix.org/', you are not a trusted user.
Run `man nix.conf` for more information on the `substituters` configuration option.
error:
       … while calling the 'derivationStrict' builtin

         at //builtin/derivation.nix:9:12: (source not available)

       … while evaluating derivation 'ghc-shell-for-hackage-server-0.5.1-0'
         whose name attribute is located at /nix/store/v798823rqbsn3qx22rwayvcbiz5gn89x-source/pkgs/stdenv/generic/make-derivation.nix:300:7

       … while evaluating attribute '__impureHostDeps' of derivation 'ghc-shell-for-hackage-server-0.5.1-0'

         at /nix/store/v798823rqbsn3qx22rwayvcbiz5gn89x-source/pkgs/stdenv/generic/make-derivation.nix:433:7:

          432|       __propagatedSandboxProfile = lib.unique (computedPropagatedSandboxProfile ++ [ propagatedSandboxProfile ]);
          433|       __impureHostDeps = computedImpureHostDeps ++ computedPropagatedImpureHostDeps ++ __propagatedImpureHostDeps ++ __impureHostDeps ++ stdenv.__extraImpureHostDeps ++ [
             |       ^
          434|         "/dev/zero"

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: assertion '(stdenv).isLinux' failed

       at /nix/store/v798823rqbsn3qx22rwayvcbiz5gn89x-source/pkgs/os-specific/linux/kernel/generic.nix:71:1:

           70|
           71| assert stdenv.isLinux;
             | ^
           72|
```
So this does not seem to work on MacOS after all, or am I doing something wrong?  I am a total nix nitwit, so maybe I am missing something otherwise obvious.
I am on `nix 2.15.0`.

cabal.project
```diff
@@ -1,3 +1,9 @@
+packages:
+         .
```
andreasabel, line 2: Does the `.` have to be on the following line to enable parsing this by `flake.nix`?

(Anyway, `packages: .` is the default so I don't know why it is here in the first place.  Could maybe be deleted, but then posterity could add it again in a way that `flake.nix` does not understand...)

andreasabel (commented):
Thanks!

Is this supposed to work on macOS?  ("this" is e.g. `nix develop`)

README.md
peterbecich: Good catch, thanks.

I forgot to apply this fix https://github.com/haskell/hackage-server/pull/1219

peterbecich: This comment specifically https://github.com/haskell/hackage-server/pull/1219#issuecomment-1597140858

cabal.project
peterbecich: Yes, I believe it is a bug https://github.com/srid/haskell-flake/issues/193

andreasabel: Seems like the parser there wants this layout:
```
packages:
  foo
  bar
```

peterbecich: Interestingly I think the error still occurs with only two spaces:
```
packages:
  .
```
I checked yesterday.

peterbecich: improved, it works with two spaces, thanks

README.md
peterbecich: @andreasabel , please test 
```
nix develop  --extra-experimental-features nix-command --extra-experimental-features flakes
```
again on branch `master`.

The best solution is for the GitHub Action to check `nix develop`. This would check both Mac and Linux. However I don't know if that is possible.

peterbecich: I merged the PR because apparently it was already broken for Mac, so merging it would at worst leave it broken.

peterbecich: Okay, I think this reproduces the issue in a GitHub Action:

https://github.com/haskell/hackage-server/actions/runs/5946853679/job/16128037931?pr=1220

`nix flake check` checks `nix develop`, as @\srid explained to me here: https://github.com/srid/haskell-flake/discussions/172#discussioncomment-6219087

-----------

No, that warning message is not the cause.

If you want to remove that warning on your Mac, you need to edit the root-owned `nix.conf`. There are two `nix.conf`. One owned by root, and another by your user.

Here is the location of the former on Mac:  https://stackoverflow.com/a/63150105/1007926

You can add these settings:
```
trusted-users = [whoami username]

experimental-features = nix-command flakes
```

https://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-trusted-users

Then restart the Nix daemon; unsure how to do on Mac. Or restart the whole system.

Issue #1247, opened 2023-08-26
Nix Flake `develop` broken for Mac


peterbecich:
We need to figure out why `nix develop` on Mac pulls in Linux dependencies.

https://github.com/haskell/hackage-server/blob/f52a15d3fdc82f9e8b9368e7c85d288836eda3e0/flake.nix#L34-L57

Error in `nix develop` in Mac:
https://github.com/haskell/hackage-server/pull/1244#discussion_r1301157439
```

           70|
           71| assert stdenv.isLinux;
             | ^
           72|
```

The error can also be seen in `nix flake check` on the Mac GitHub Action:
https://github.com/haskell/hackage-server/actions/runs/5982632728/job/16231916472

`nix flake check` checks `nix develop`: https://github.com/srid/haskell-flake/discussions/172#discussioncomment-6219087

peterbecich:
We should also check `--show-trace`: https://github.com/srid/haskell-flake/discussions/172#discussioncomment-6224814

peterbecich:
We should also check `--show-trace`: https://github.com/srid/haskell-flake/discussions/172#discussioncomment-6224814

peterbecich:
Should be fixed, @andreasabel , please try `nix develop` again on your Mac

andreasabel:
Should be ok.  I interrupted it though as it was downloading GBs, and I got a warning that my harddrive was running full, so I cured it with `sudo rm -rf /nix/store`.

PR #1248, opened 2023-08-26
fix `nix develop` on Mac

peterbecich:
https://github.com/haskell/hackage-server/issues/1247

https://github.com/srid/haskell-flake/discussions/172#discussioncomment-6224814


peterbecich:
The issue was I left the dependency `cryptodev` in the `devShell` config

PR #1256, opened 2023-09-23
Close <p> tags in the revisions RSS feed

Kleidukos:


bgamari:
For my own reference, is it actually necessary to close these? My impression was that HTML5 did not require closing tags for `<p>` and `<li>` (in contrast to XHTML).

Kleidukos:
@bgamari I don't believe we specify HTML5 in the rss feed's content. Either way, I don't think we should juggle between using xhtml (which is a dependency of hackage) and HTML5 when we don't feel like closing tags. Consistency is best.

Issue #1261, opened 2023-10-26
Unique filenames for `$revision.cabal`.


lockywolf:
Dear Haskell developers,

At the moment, versioned `.cabal` files are shipped under the name of `$revision.cabal`, e.g., `1.cabal`. This makes it impossible to download many cabal files into a directory, because of the name clash.

Please, consider adding an url for the file with the name `package-$revision.cabal`.

Example:
In addition to having `https://hackage.haskell.org/package/network-uri-2.6.4.2/revision/1.cabal`, also add `https://hackage.haskell.org/package/network-uri-2.6.4.2/revision/network-uri-2.6.4.2-1.cabal`.

lockywolf:
I think that something like that should be basically enough
```
diff --git a/datafiles/templates/Html/revisions.html.st b/datafiles/templates/Html/revisions.html.st
index 69cf2b75..f499d0ea 100644
--- a/datafiles/templates/Html/revisions.html.st
+++ b/datafiles/templates/Html/revisions.html.st
@@ -25,7 +25,7 @@ stored separately.
   </tr>
   $revisions:{revision|
     <tr>
-      <td valign="top"><a href="/package/$pkgid$/revision/$revision.number$.cabal">-r$revision.number$</a></td>
+      <td valign="top"><a href="/package/$pkgid$/revision/$revision.number$.cabal">-r$revision.number$</a>(<a href="/package/$pkgid$/revision/$pkgid$-$revision.number$.cabal">$pkgid$-r$revision.number$</a>)</td>
       <td valign="top">$revision.htmltime$</td>
       <td valign="top"><a href="/user/$revision.user$">$revision.user$</td>
       <td valign="top">$revision.sha256$</th>
diff --git a/src/Distribution/Server/Features/Core.hs b/src/Distribution/Server/Features/Core.hs
index 6aec793a..b940248d 100644
--- a/src/Distribution/Server/Features/Core.hs
+++ b/src/Distribution/Server/Features/Core.hs
@@ -455,6 +455,9 @@ coreFeature ServerEnv{serverBlobStore = store} UserFeature{..}
     coreCabalFileRev = (resourceAt "/package/:package/revision/:revision.:format") {
         resourceDesc = [(GET, "Get package .cabal file revision")]
       , resourceGet  = [("cabal", serveCabalFileRevision)]
+    coreCabalFileRev = (resourceAt "/package/:package/revision/:package-:revision.:format") {
[...]

PR #1262, opened 2023-10-26
Allow serving cabal files with package name in the file name.

lockywolf:
Allow serving url like this:
https://hackage.haskell.org/package/pkg-2.6/revision/pkg-2.6-1.cabal

Resolves: https://github.com/haskell/hackage-server/issues/1261 #1261

### DIFF
--- a/datafiles/templates/Html/revisions.html.st
+++ b/datafiles/templates/Html/revisions.html.st
@@ -25,7 +25,7 @@ stored separately.
   </tr>
   $revisions:{revision|
     <tr>
-      <td valign="top"><a href="/package/$pkgid$/revision/$revision.number$.cabal">-r$revision.number$</a></td>
+      <td valign="top"><a href="/package/$pkgid$/revision/$revision.number$.cabal">-r$revision.number$</a> (<a href="/package/$pkgid$/revision/$pkgid$-$revision.number$.cabal">$pkgid$-r$revision.number$</a>)</td>
       <td valign="top">$revision.htmltime$</td>
       <td valign="top"><a href="/user/$revision.user$">$revision.user$</td>
       <td valign="top">$revision.sha256$</th>
--- a/src/Distribution/Server/Features/Core.hs
+++ b/src/Distribution/Server/Features/Core.hs
@@ -225,6 +225,7 @@ data CoreResource = CoreResource {
     corePackageTarball  :: Resource,
     -- | A Cabal file metatada revision.
     coreCabalFileRev    :: Resource,
+    coreCabalFileRevName    :: Resource,
 
     -- Rendering resources.
     -- | URI for `corePackagesPage`, given a format (blank for none).
@@ -403,6 +404,7 @@ coreFeature ServerEnv{serverBlobStore = store} UserFeature{..}
           , coreCabalFile
           , coreCabalFileRevs
           , coreCabalFileRev
+          , coreCabalFileRevName
           , coreUserDeauth
           , coreAdminDeauth
           , corePackUserDeauth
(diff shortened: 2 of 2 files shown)


datafiles/templates/Html/revisions.html.st
```diff
@@ -25,7 +25,7 @@ stored separately.
   </tr>
   $revisions:{revision|
     <tr>
-      <td valign="top"><a href="/package/$pkgid$/revision/$revision.number$.cabal">-r$revision.number$</a></td>
+      <td valign="top"><a href="/package/$pkgid$/revision/$revision.number$.cabal">-r$revision.number$</a>(<a href="/package/$pkgid$/revision/$pkgid$-$revision.number$.cabal">$pkgid$-r$revision.number$</a>)</td>
```
phadej, line 28: There probably should be a space between opening parenthesis?

src/Distribution/Server/Features/Core.hs
```diff
@@ -456,6 +458,11 @@ coreFeature ServerEnv{serverBlobStore = store} UserFeature{..}
         resourceDesc = [(GET, "Get package .cabal file revision")]
       , resourceGet  = [("cabal", serveCabalFileRevision)]
       }
+    coreCabalFileRevName = (resourceAt "/package/:package/revision/:package-:revision.:format") {
```
phadej, line 461: I wonder how `:package...:package` template works. It probably doesn't (i.e. allows non-sense like `/package/foo/revision/bar-1-r2.cabal`, or even `/package/non---sense-1-1-1-1/revision/bar-1-r2.cabal`

See e.g

```
    corePackageTarball = (resourceAt "/package/:package/:tarball.tar.gz") {
   --- which uses:
    packageTarballInPath dpath = do
      PackageIdentifier name version <- packageInPath dpath
      case lookup "tarball" dpath >>= fromReqURI of
        Nothing -> mzero
        Just pkgid@(PackageIdentifier name' version') -> do
          -- rules:
          -- \* the package name and tarball name must be the same
          -- \* the tarball must specify a version
          -- \* the package must either have no version or the same version as the tarball
          guard $ name == name' && version' /= nullVersion && (version == version' || version == nullVersion)
          return pkgid
```

lockywolf: Why is that a problem? If someone is using a malformed url, they will get malformed data. I.e. it's their problem.

Is there a security issue here I don't see?

phadej: > Is there a security issue here I don't see?

There is CDN in from of hackage-server which caches the results. I'm sure someone can be annoying, as there would be infinitely many cacheable urls. Maybe not a security risk, but in my opinion wrong URLs should be reported as the wrong ones.

lockywolf: Okay. Then I didn't know how to implement it. I can monkey-copy the code that closely matches someone else's, but I don't know how to implement unification in Haskell.

In the defence of my, undoubtedly, flawed, position, I can say that github also allows infinitely many cacheable URLs, like https://github.com/username/project/archive/v1.0.0/bar-$version.tag.gz

phadej: > Then I don't know how to implement it.

I gave you a pointer. Look how `corePackageTarball` is implemented (using `packageTarballInPath` which you can use as an example).

lockywolf: Please theck the new version. I have reused the "tarball" function, because, as far as I understand, it is essentially the same.

But maybe I misunderstood.

lockywolf: @phadej does the new version make sense? Although I re-used the function with "tar" in the name, it seems that the pattern match is the same.

lockywolf:
What do the failures with those two older compilers mean?

PR #1267, opened 2023-11-29
Add hashes to the list of revisions in JSON format

andreabedini:
The list of metadata revisions for a package is available at
the url /package/:package/revisions/

When rendered in HTML this list includes the revision number, timestamp,
uploader, and sha256; but when rendered in JSON the sha256 is missing.

This change adds revisions sha256 to the JSON rendering.

This simple change be of help in situations where you need to retrieve a cabal file by its hash. Lacking a form or content-addressing, this allows finding the cabal file by the knowlege of its hash and package-id in two calls.

### DIFF
--- a/src/Distribution/Server/Features/Core.hs
+++ b/src/Distribution/Server/Features/Core.hs
@@ -24,7 +24,7 @@ module Distribution.Server.Features.Core (
 
 -- stdlib
 import qualified Codec.Compression.GZip                             as GZip
-import           Data.Aeson                                         (Value (..))
+import           Data.Aeson                                         (Value (..), toJSON)
 import qualified Data.Aeson.Key                                     as Key
 import qualified Data.Aeson.KeyMap                                  as KeyMap
 import           Data.ByteString.Lazy                               (ByteString)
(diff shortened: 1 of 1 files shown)


src/Distribution/Server/Features/Core.hs
```diff
@@ -728,12 +729,15 @@ coreFeature ServerEnv{serverBlobStore = store} UserFeature{..}
...
-            let uname = userIdToName users uid in
+          revisionToObj rev (cabalFileText, (utime, uid)) =
+            let uname = userIdToName users uid
+                hash = sha256 (cabalFileByteString cabalFileText)
+            in
             Object $ KeyMap.fromList
               [ (Key.fromString "number", Number (fromIntegral rev))
               , (Key.fromString "user", String (Text.pack (display uname)))
               , (Key.fromString "time", String (Text.pack (formatTime defaultTimeLocale "%Y-%m-%dT%H:%M:%SZ" utime)))
+              , (Key.fromString "sha256", String (Text.pack (show hash)))
```
ysangkok, line 740: Since this type has [a `ToJSON` instance](https://github.com/haskell/hackage-server/blob/e37b78bf3137d3e45852e87904d9120e9ee31dd2/src/Distribution/Server/Features/Security/SHA256.hs#L70), I think it would be better to use that instead of Show. As the behaviour of Show isn't very precisely defined.

PR #1268, opened 2023-11-30
fix: replace list lastVersion by referenceVersion (#1264)

### LINKED ISSUE #1264: Hackage search showing package version inconsistency
When using the search on hackage, the resulting package list can show the incorrect package version (apparently it shows the most recent uploaded package).

I noticed that behaviour with the BNFC package. Recently the version 2.9.5 was uploaded, and after that version 2.5.0.1 was also uploaded.

The problem is that searching for BNFC now list the last version as 2.5.0.1, but links correctly for the (last) default version (2.9.5).
This is not really serious issue , but is a bit confusing...

blackheaven:
An attempt to fix #1264

### DIFF
--- a/datafiles/static/browse.js
+++ b/datafiles/static/browse.js
@@ -133,7 +133,7 @@ const replaceRows = (response) => {
     tr.appendChild(createSimpleText(row.description));
     tr.appendChild(createTags(row.tags));
     tr.appendChild(createLastUpload(row.lastUpload));
-    tr.appendChild(createSimpleText(row.lastVersion));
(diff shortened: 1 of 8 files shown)


blackheaven:
Ok, it's ready to be reviewed

blackheaven:
I have tested with an existing package and it worked.

I'll try your case today.

I'll also try to make it retroactive, I was reluctant because it involved an important refactoring, but, anyway...

blackheaven:
It was tedious but it's retroactive, I have tested it, hopefully I have not missed another case.

Issue #1218, opened 2023-06-18
UI for multiple public libraries


michaelpj:
Looks like there currently isn't any, really. https://hackage.haskell.org/package/lsp-types has a second public library but there's no evidence of it on the page.

gbaz:
mildly related https://github.com/haskell/hackage-server/issues/577

ysangkok:
I can work on this after I get vouching done.

414owen:
I'm happy to take a stab at this.

I've made changes that get the modules to show per-library. Seems to work fine.

I'd also be in favour of listing dependencies per-library on the main package page (as opposed to having to click into the easy-to-miss `details` section). One of the use-cases of sub-libraries it to have a lightweight core library, then separate sub-libraries for integrating into other parts of the ecosystem. For instance, I have a library that depends only on `base`, and a sub-library which adds instances from the `semigroupoids` package. I'd like to make it clear that my (main) library is lightweight.

Not sure if the view heirarchy should be:

* Modules
  * lib-a
  * lib-b
* Sigs
  * lib-a
  * lib-b

or

* lib-a
  * Modules
  * Sigs
* lib-b
  * Modules
  * Sigs


I'll play around with it and see what feels good.

gbaz:
Great to hear you're working on the changes!

I think the second approach is better, with the caveat that we shouldn't have the "lib-a, lib-b" top level package hierarchy when there is only one exposed public library. Similarly we shouldn't have the signatures section and modules section split in the second level when there are only modules.

I'm sympathetic to wanting to have the dependencies per-library. However, for packages with many dependencies, splitting them out would make things even more cluttered than they are now. Maybe we can just highlight the details link better, and perhaps make it more descriptive?

414owen:
Yep, that's all sensible. I'll optimise for lack of visual clutter.

If I figure out how to display the dependencies per-library in an uncluttered way, it'll be in a separate changeset anyway.

michaelpj:
The current approach seems to be to do something like: for each top-level section, split it up by component.

This seems fine for packages with a smallish number of components, but I'm not sure it's going to scale. In particular, imagine if `amazonka` actually did move over to multiple public libraries. Then there would be dozens of libraries! It would then be really annoying trying to cross-reference the dependencies (2/3 of the way down one massive element) with the haddock (2/3 of the way down a different massive element).

[...]

PR #1279, opened 2023-12-30
feat: Add support for sublibrary module listings

414owen:
works on https://github.com/haskell/hackage-server/issues/1218

## Changes:

* Adds a "renderer" for library names
* Uses the library name renderer in the dependency details page
* Add module and signature trees per-library on the package page

I figured it's appropriate to make sure libraries are shown by the same name in the dependency-details page, and on the package page, hence the library name renderer. I also think it's more intuitive to refer to the libraries by the same name that is used to add them as dependencies, rather than just saying `library`, or using just the sublibrary component name.

## Screenshots

Using the unpacked-containers package, as it has sublibraries, and signatures as well as modules

### Before

Screenshot from 2023-12-30 20-18-28
Screenshot from 2023-12-30 19-47-05

### After

Screenshot from 2023-12-30 20-18-20

Screenshot from 2023-12-30 19-46-56

### DIFF
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -1146,6 +1146,14 @@ a.deprecated[href]:visited {
     color: #61B01E;
 }
 
+.lib-contents {
+  margin-left: 20px;
+}
+
+.lib-contents > h3 {
+  margin: 0.7em 0;
+}
+
 /* Paginator */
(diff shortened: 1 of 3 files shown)


gbaz (approved):
very cleanly done, thanks!

PR #1280, opened 2024-01-01
feat: Specify component types in detailed deps page

414owen:
With the advent of multiple sublibraries per
package, it has become less clear what each
component is, on the 'detailed dependencies'
page.

This changeset replaces the tabulated view of

| component name | list of dependencies |
| -------------- | -------------------- |
| - lib 1        | - dep 1              |
|                | - dep 2              |
| - exe 1        | - dep 3              |

With the more heirarchical:

Libraries
  - lib 1
    - dep 1
    - dep 2 ...

Executables
  - exe 1
    - dep 3

This change is related to https://github.com/haskell/hackage-server/issues/1218

---

### Before

Screenshot from 2024-01-01 23-04-03

### After

Screenshot from 2024-01-01 23-01-48

Obviously, when there are no components of a given type ({Executables, Libraries}), the header itself is not rendered.


ysangkok:
Thank you, this seems useful.

PR #1283, opened 2024-01-07
Migrate to tar-0.6

### LINKED ISSUE #1277: Support tar-0.6
`hackage-server` demands `tar < 0.6`, prohibiting a recently released https://hackage.haskell.org/package/tar-0.6.0.0.

Bodigrim:
Closes #1277.

### DIFF
--- a/src/Distribution/Server/Packages/Unpack.hs
+++ b/src/Distribution/Server/Packages/Unpack.hs
@@ -138,10 +138,10 @@ tarPackageChecks lax now tarGzFile contents = do
       expectedDir = display pkgid
 
       selectEntry entry = case Tar.entryContent entry of
-        Tar.NormalFile bs _         -> Just (normalise (Tar.entryPath entry), NormalFile bs)
-        Tar.Directory               -> Just (normalise (Tar.entryPath entry), Directory)
-        Tar.SymbolicLink linkTarget -> Just (normalise (Tar.entryPath entry), Link (Tar.fromLinkTarget linkTarget))
-        Tar.HardLink     linkTarget -> Just (normalise (Tar.entryPath entry), Link (Tar.fromLinkTarget linkTarget))
+        Tar.NormalFile bs _         -> Just (normalise (Tar.entryTarPath entry), NormalFile bs)
+        Tar.Directory               -> Just (normalise (Tar.entryTarPath entry), Directory)
+        Tar.SymbolicLink linkTarget -> Just (normalise (Tar.entryTarPath entry), Link linkTarget)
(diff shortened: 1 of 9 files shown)


Bodigrim:
Could someone help me out with Nix Flake please? Never used Nix in my life unfortunately.

peterbecich:
Fixing it, will append a commit to this PR, sorry for the inconvenience

peterbecich:
I think it is fixed. Pardon the `git rebase`. It was fixed by `nix flake update` and other tweaks to the Flake so `flake update` would succeed:  https://github.com/haskell/hackage-server/pull/1285 Your addition to the Flake `tar.source = "0.6.0"` was correct; the `flake` was too far behind NixPkgs for this to work, however.

Bodigrim:
Thanks @peterbecich! 

@gbaz @ysangkok @andreasabel this is ready for review.

src/Distribution/Server/Packages/Unpack.hs
```diff
@@ -468,6 +480,10 @@ explainTarError (PermissionsError entryname mode) =
   where
     showMode :: Tar.Permissions -> String
     showMode m = printf "%.3o" (fromIntegral m :: Int)
+explainTarError (LongNamesError err) =
+    "There is an error in the format of entries with long names in the tar file: " ++ show err
+ ++ ". Check that it is a valid tar file (e.g. 'tar -xtf thefile.tar'). "
+ ++ "You may need to re-create the package tarball and try again."
```
andreasabel, line 486: Can we get a test case that triggers this error?  

Could be not easy to engineer, but it is good to have evidence for every error that it is possible to trigger.

andreasabel (commented):
Thanks for the upgrade!

src/Distribution/Server/Packages/Unpack.hs
Bodigrim: Are there tests for pre-existing errors somewhere? I can add a new one there.

This is a rare condition though: one has to have an intent to craft a malicious TAR, you cannot obtain one by mistake.

andreasabel: > Are there tests for pre-existing errors somewhere? I can add a new one there.

Yeah, that is the usual problem: There aren't enough tests to start with, so there isn't a good pretext to add some for new or changed features.  One can't require adding a test infrastructure for a maintenance PR like this.

andreasabel (approved):

Bodigrim:
@gbaz @ysangkok any more reviews? Or are you happy to merge?

ysangkok:
Looks ok to me. Happy to merge.

PR #1292, opened 2024-02-06
feature: add 'reporting vulnerabilities' link on package pages

blackheaven:
It's the implementation of [haskell/security-advisories#7](https://github.com/haskell/security-advisories/issues/7)

It looks like this:

image


frasertweedale:
I think it's a fine first step.  Over time it would be nice to have a more user-friendly process (e.g. submit a form, rather than GitHub pull request), but this is better than nothing!

I suggest changing the text to *"Report a vulnerability"* or *"Report security issues"*.  i.e. imperative tense to conform with *"Propose tags"*.

blackheaven:
I have pushed the suggested change.

Sure, we can draft it further on the original issue.

PR #1308, opened 2024-04-07
Add Tested-With to package page

ysangkok:
Is this what you wanted @telser ?

Preview with smtp-mail:

smtp-mail side pane


telser:
This is great, certainly helps to see what compilers a package is targeting.

Issue #1298, opened 2024-03-12
quotes have very inaccessible / illegible styling 


MangoIV:
![image](https://github.com/haskell/hackage-server/assets/40720523/78ad5bb1-ea79-493e-a77e-9282e55659ee)

see e.g. https://hackage.haskell.org/api

PR #1319, opened 2024-06-22
fix quote color

peterbecich:
closes: https://github.com/haskell/hackage-server/issues/1298

image

### DIFF
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -270,7 +270,7 @@ pre + pre {
 @media (prefers-color-scheme: dark) {
   blockquote {
     border-left: 3px solid #2f2842;
-    background-color: #252034;
+    background-color: #4c4771;
   }
 }
 
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -267,9 +267,21 @@ pre + pre {
   margin-top: 0.5em;
 }
 
+@media (prefers-color-scheme: dark) {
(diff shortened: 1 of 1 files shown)


peterbecich:
@ffaf1 or @ysangkok, please review

ysangkok:
According to [WebAIM](https://webaim.org/resources/contrastchecker/), with the colors in the screenshot, foreground 21262F, background 281B35, the contrast is 1.06:1. I am not disabled and it's really faint for me. Of course, this isn't really foreground, and it shouldn't be too jarring, but this contrast is still really small. Another approach is too mark the quote by a line on the left, or maybe a 1em wide [45deg line pattern](https://stock.adobe.com/es/images/line-pattern-45-degree-inclined/223966159). I would be fine with a contrast of 1.5:1 or something like that. Using the lightness slider, that would mean replacing by 21262F by 49355F.

peterbecich:
Thanks, good point. The color is 252034 versus 333 but you're correct; the contrast is too low.
https://github.com/haskell/hackage-server/blob/081a1032980197487c9337526dc94daff3de6813/datafiles/static/hackage.css#L273
https://webaim.org/resources/contrastchecker/?fcolor=252034&bcolor=333333

How about 4c4771 versus 333 https://webaim.org/resources/contrastchecker/?fcolor=4C4771&bcolor=333333

image

ysangkok (approved):

PR #1322, opened 2024-07-02
Cabal 3.12

andreasabel:
This PR does the minimum changes needed for `Cabal-3.12` (and some cleanup work).

Subsumes:
- #1137

Commits:
- **Clean cabal.project**: remove some cruft
- **Default library (hackage-server) instead of internal library lib-server**: the introduction of a internal library 7 years ago never got anywhere, and internal libs still cause problems with some tooling
- **.cabal: remove outdated conditional**: remove some cruft
- **Tags feature: replace some clumsy code**: a drive-by shooting; couldn't look past this contrived and inefficient piece of code when I ran into it
- **Support Cabal-3.12 and GHC 9.10**: main commit, introducing `#if`s and the necessary code to work with latest `Cabal(-syntax)`.  Passes CI.
- **Commit to Cabal-3.12.1.0**: bump the lower bound of `Cabal(-syntax)` to latest; breaks the nix CI

I had to disable the reverse dependencies benchmark because `gauge` does not support `base-4.20` and there is no maintainer for this package dropped by @vincenth in sight.  @ysangkok, consider switiching to `criterion` or related.

Eventually we can remove all the `#if MIN_VERSION_Cabal` conditionals, but there is no rush now (they have been CI tested).

### DIFF
--- a/hackage-server.cabal
+++ b/hackage-server.cabal
@@ -149,8 +149,8 @@ common defaults
   -- other dependencies shared by most components
   build-depends:
     , aeson                  >= 2.1.0.0 && < 2.3
-    , Cabal                  >= 3.10.1.0 && < 3.14
-    , Cabal-syntax           >= 3.10.1.0 && < 3.14
+    , Cabal                  >= 3.12.1.0 && < 3.14
+    , Cabal-syntax           >= 3.12.1.0 && < 3.14
         -- Cabal-syntax needs to be bound to constrain hackage-security
         -- see https://github.com/haskell/hackage-server/issues/1130
     , fail                  ^>= 4.9.0
--- a/hackage-server.cabal
+++ b/hackage-server.cabal
@@ -28,9 +28,10 @@ license:      BSD-3-Clause
 license-file: LICENSE
 
 tested-with:
-  GHC == 9.8.1
-  GHC == 9.6.3
-  GHC == 9.4.7
+  GHC == 9.10.1
+  GHC == 9.8.2
+  GHC == 9.6.5
+  GHC == 9.4.8
   GHC == 9.2.8
   GHC == 9.0.2
   GHC == 8.10.7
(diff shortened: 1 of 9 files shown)


hackage-server.cabal
```diff
@@ -129,51 +130,52 @@ common defaults
...
     , directory              >= 1.3   && < 1.4
-    , filepath               >= 1.4   && < 1.5
+    , filepath               >= 1.4   && < 1.6
     , mtl                    >= 2.2.1 && < 2.4
         -- we use Control.Monad.Except, introduced in mtl-2.2.1
     , pretty                 >= 1.1   && < 1.2
     , process                >= 1.6   && < 1.7
     , text                  ^>= 1.2.5.0 || >= 2.0 && < 2.2
-    , time                   >= 1.9   && < 1.13
+    , time                   >= 1.9   && < 1.15
```
ysangkok, line 145: Was time-1.14 actually tested? time is different from other boot libraries because GHC-9.10 doesn't actually use the newest version. So e.g. assuming that that solving all of `cabal outdated` is necessary for GHC-9.10 is wrong. GHC-9.10 still uses time-1.12. I don't have an issue with bumping this, but then we need to make sure we're testing it.

peterbecich:
Is `Cabal-3.12.0.0` sufficient? This is already available in `nixpkgs`. I am trying to update `nixpkgs` to 3.12.1.0 here: https://github.com/NixOS/nixpkgs/pull/324424

Bodigrim:
Once the upgrade to `Cabal-3.12` is merged, please merge https://github.com/haskell/hackage-server/pull/1137 as well.

hackage-server.cabal
andreasabel: No wasn't tested, I can go back to 1.12.

andreasabel:
@peterbecich wrote:
> Is `Cabal-3.12.0.0` sufficient? This is already available in `nixpkgs`. I am trying to update `nixpkgs` to 3.12.1.0 here: [NixOS/nixpkgs#324424](https://github.com/NixOS/nixpkgs/pull/324424)

Unfortunately, `Cabal(-syntax)-3.12.1.0` adds the `MHS` constructor (micro Haskell), breaking the PVP:
- https://github.com/haskell/cabal/issues/10161

So either we commit to `Cabal >= 3.12.1.0` or to `Cabal == 3.12.0.*`.  The former is more future compatible so I opted for it.

andreasabel:
@Bodigrim wrote:
> Once the upgrade to `Cabal-3.12` is merged, please merge #1137 as well.

Thanks for the pointer!
I should take the missing parts of #1137 and include it into this PR.  (This is just the `arbitrary` implementation for architecture values.)
- #1137

hackage-server.cabal
andreasabel: Restricted to `time < 1.13` again.

ysangkok (approved):

andreasabel:
> I should take the missing parts of #1137 and include it into this PR.

Done.

Bodigrim:
It would be great if upgrade to Cabal 3.12 could be deployed soon. Is there anything blocking it?

ysangkok:
> Cabal 3.12 could be deployed soon. Is there anything blocking it?

We'd have to ask @gbaz , I think.

PR #1327, opened 2024-08-04
use GHC 9.8 in Nix Flake

peterbecich:


ysangkok:
I think it would cause less churn to start using GHC 9.8 only after Nixpkgs has upgraded to a Stackage LTS with GHC 9.8.

PR #1344, opened 2024-11-22
Detailed doc string on /package/{name}.json

ysangkok:
I want to make the Renovate pull version information from Hackage. It would be
nice to use this endpoint, and have Hackage commit to the API having a certain
structure; which it could do by documenting the response body.


MangoIV (approved):
I think this is a good improvement.

PR #1348, opened 2024-11-24
fix expected content-types for documentation tarballs

### LINKED ISSUE #1269: Allow uploading docs as tar.gz
When making `cabal haddock --haddock-for-hackage`, the `.tar.gz` package is produced.

If I upload it manually to Hackage,

> I get For this resource the content-type was expected to be application/x-tar, rather than application/gzip

which is quite inconvenient

MangoIV:
Fixes #1269

- documentation tarballs produced by cabal haddock are compressed
- their mimetype is application/gzip
- keeps the applicatoin/x-tar and applicatoin/x-gzip even though there is no tar mimetype and there's now (since 2014) a gzip mimetype, according to RFC6713
- remove the expectUncompressedTarball function as it is now dead code
- remove a pair of redundant paren and replace infix `liftM` with <$>


MangoIV:
Pinging @phadej and @gbaz for review as they were part of the original issue. 

Thank you.

ulysses4ever:
@gbaz I believe this would be worth merging

PR #1355, opened 2025-01-03
Mangoiv/more space

MangoIV:


ulysses4ever:
Yes, please!

In the "previously" you have screenshots for both desktop and mobile but for "now" it's only mobile. Is it possible to get desktop too?

MangoIV:
> Is it possible to get desktop too?

Yesterday GitHub didn’t wanna :P 

I can try again today

ulysses4ever:
sounds good. Keep us posted about that, please!

src/Distribution/Server/Pages/Package.hs
```diff
@@ -95,7 +95,7 @@ downloadSection PackageRender{..} =
             [ anchor ! [href cabalURL] << "Package description" ]
           (True,Nothing) ->
             [ anchor ! [href cabalURL] << "Package description"
-            , toHtml " (as included in the package)"
+            , toHtml " (as included in the package as *.cabal file)"
```
gbaz, line 98: this isn't great grammar. "as included in the package cabal file" would be fine.

datafiles/templates/Html/package-page.html.st
```diff
@@ -283,9 +260,37 @@
               <th>Status</th>
               <td>$buildStatus$</td>
             </tr>
+            <tr>
```
gbaz, line 263: I'm ok with moving to sidebar, but do not want it hidden by default. all the info except "edit package information" is generally relevant -- checking who the maintainers are, what the candidates are, etc.

MangoIV:
@ulysses4ever whats your opinion on hiding maintainers corner by default? If there are no other opinions I’ll switch back to not hiding by default like @gbaz requested…

ulysses4ever:
I often seek the list of maintainers as defined by Hackage, not the cabal file, so I'd definitely not want to be "hidden" by default. Sidebar is a good place, I think.

Candidates are sometimes relevant, so I'd lean towards not hiding them either.

I don't know why either of these two (maintainers list or candiadtes) are under "Mainteiner's corner". I'd prefer them to be in the list of package features in the sidebar.

Having the "Edit" link or the title "Maintainer's corner" visible to anyone other than maintainers I find unfortunate. But that's a consequence of how the ancient authentication mechanism prompts you to go about it, I guess: there's no explicit Login button (easily available?) and the easiest way to log in is to click the Edit link. 

Now is the time to ask you @MangoIV to provide a screenshot of what it means to be "hidden" in your patch, because I'm bad at compiling HTML in my head, and a mockup would help me to decide whether I prefer "Edit" to be "hidden". In any case, if it's pretty much one link at this point, there's no reason to split hairs over this: I'd just go with what Gershom suggests.

MangoIV:
Ah perfect, that’s a good helper to enforce that decision. I’ll also put up a screenshot when I get around to it. But I think both of your arguments are fair!

MangoIV:
I'm still not entirely happy with how this looks, especially on pages with very large module lists. I think the README should come first and then both the modules and the rest should be side bars? But that's also not idea because that requires a lot of width.

ulysses4ever:
Thanks for the screenshots!

maybe we need tabs like Rust’s crates.io?.. 

They also have less busy right sidebar with metadata, which seems like a good idea. Dependencies, one of the bigger fields in our sidebar, go in a tab with their design.

E.g. https://crates.io/crates/syn

IMG_0589

MangoIV:
yeah I think that would be really nice.

PR #1375, opened 2025-04-01
Add uploader's username to package json API

Kleidukos:
This gives the following payload:

```diff
{
    "author": "qwbarch",
    "copyright": "2021 qwbarch",
    "description": "Haskell package for easy integration with the 2captcha API.\n\nFeature list:\n\n* Lens-based API\n* Uses Wreq for http calls",
    "homepage": "https://github.com/qwbarch/2captcha-haskell#readme",
    "license": "MIT",
    "metadata_revision": 0,
    "synopsis": "Haskell package for easy integration with the 2captcha API.",
    "uploaded_at": "2022-09-24T05:49:45Z",
+    "uploader": "admin"
}
```

### DIFF
--- a/src/Distribution/Server/Features.hs
+++ b/src/Distribution/Server/Features.hs
@@ -379,6 +379,7 @@ initHackageFeatures env@ServerEnv{serverVerbosity = verbosity} = do
     packageInfoJSONFeature <- mkPackageJSONFeature
                                 coreFeature
                                 versionsFeature
+                                usersFeature
 
 #endif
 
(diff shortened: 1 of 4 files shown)


gbaz:
@Kleidukos I shouldn't have merged this. Testing reveals that it doesn't work, because it changes the statefile without creating a migration. But that said, why is there state at all? Can't we just have a lazy in-memory cache with no persistence? A refactor to do that would probably be easier, and appreciated.

Kleidukos:
@gbaz oh shit! I thought CI would have caught that. Of course locally it all worked perfectly…

What tools do we have at our disposal in the app for lazy in-memory caching? I must admit I am very unfamiliar with the backend's subtleties.

gbaz:
I mean literally just don't store this stuff in acid-state. store it in a map, and query it from the map.

gbaz:
on CI, yeah, we don't test from a dirty state, which would be a bit of engineering, but welcome, since it would catch this stuff!

Kleidukos:
> I mean literally just don't store this stuff in acid-state. store it in a map, and query it from the map.

Aaah I see. :)

PR #1385, opened 2025-04-19
BuildClient: Port to http-client-tls

bgamari:
This introduces TLS support in the documentation builder by porting the HTTP abstraction in `Distribution.Client` from `HTTP` to `http-client-tls`. This is currently deployed on the new documentation builder.

While in the area I also do a bit of housekeeping, dropping some dead code and resolving a few warnings.

### DIFF
--- a/hackage-server.cabal
+++ b/hackage-server.cabal
@@ -409,6 +409,9 @@ library
   build-depends:
     , HStringTemplate       ^>= 0.8
     , HTTP                  ^>= 4000.3.16 || ^>= 4000.4.1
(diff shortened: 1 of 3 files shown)


hackage-server.cabal
```diff
@@ -454,6 +457,7 @@ library
     , stm                   ^>= 2.5.0
     , stringsearch          ^>= 0.3.6.6
     , tagged                ^>= 0.8.5
+    , transformers          ^>= 0.6
```
andreasabel, line 460: That's hardly compatible with older GHCs, so this broke the Haskell CI.

PR #1386, opened 2025-04-19
github-ci: Regenerate

bgamari:
Moves to Ubuntu 24.04, fixing broken jobs.


andreasabel:
That didn't work, did it?

PR #1394, opened 2025-05-12
Move uploader out of acid-state

Kleidukos:
Follow-up of #1375 to use a Map rather than storing usernames in acid-state.


Kleidukos:
@gbaz I'd be grateful if you could help me put the last pieces of the puzzle together. :)

gbaz:
ok i think i see the confusion. what i mean is why do we need any acid-state for any of this? can't we just compute it on the fly? before i had suggested a better way to cache, but i don't think we need to cache it either. hitting these endpoints should be rare, and all the data is available to query elsewhere. I think we shouldn't need acid-state specific to back this json endpoint at all, much less to migrate it.

gbaz:
Wouldn't that subsume this?

Kleidukos:
@gbaz Cheers. :) I wasn't sure how deep we could clean this up and not get bitten in prod. :)

PR #1400, opened 2025-05-20
Require user content domain for Package Candidates and Package Contents

ysangkok:
This is another follow-up to #1379.


ysangkok:
I merged this in a way that wasn't detected by Github. The commit ID is the same on master: 33d9ea3de9a298b21b53ba23386af2742384e627

PR #1404, opened 2025-06-10
update Flake

peterbecich:
`nix flake update`


peterbecich:
the flake is already failing to build on `master` so there is no downside to merging this

I have tested `nix develop` and `nix build` on my machine

PR #1391, opened 2025-05-11
Bump to Cabal 3.14 and GHC 9.12

andreasabel:
This commits to Cabal-3.14 and enables build with GHC 9.12.
Things just compiled fine with Cabal-3.14.

However, there might be accommodations necessary for Cabal-3.14.

The nix flake needs update, of course.

PR #1408, opened 2025-06-13
use nothing-but-nix to give GitHub Action more disk

peterbecich:
https://github.com/wimpysworld/nothing-but-nix

https://github.com/haskell/hackage-server/pull/1391#issuecomment-2957288948


peterbecich:
https://github.com/wimpysworld/nothing-but-nix/issues/28

PR #1413, opened 2025-07-06
Document effect of PR #761 (2018-06-25)

mpilgrem:
See:
* #761
* #1412 (the equivalent for the `master` branch)

Also documents what is documented in Cabal's documentation about what Hackage will accept and reject regarding 'the ASCII subset'. See the note at https://cabal.readthedocs.io/en/stable/cabal-package-description-file.html#package-properties


gbaz:
now live on hackage.

PR #1419, opened 2025-07-26
Support Cabal-3.16

Bodigrim:


src/Distribution/Server/Framework/Instances.hs
```diff
@@ -250,7 +250,9 @@ instance SafeCopy CompilerFlavor where
     putCopy JHC               = contain $ putWord8 7
     putCopy LHC               = contain $ putWord8 8
     putCopy UHC               = contain $ putWord8 9
+#if !MIN_VERSION_Cabal_syntax(3,16,0)
```
ulysses4ever, line 253: Wouldn't it make sense to drop sorry for HaskellSuite as well?

ulysses4ever: > There is no mention of HaskellSuite removal in https://github.com/haskell/cabal/blob/master/release-notes/Cabal-3.16.0.0.md. Is that a mistake in code or changelog?

I see: 

> Remove knowledge about hmake and haskell-suite https://github.com/haskell/cabal/pull/10912

phadej: @ulysses4ever that's a breaking change in library API. It's significant change.

ulysses4ever: Fair point, thank you! Will try to adjust the process so that we get all API changes marked as significant.

exes/Main.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE PatternSynonyms #-}
```
andreasabel, line 2: Shouldn't this rather be added to the `default-extensions` in the `common defaults` section of the `.cabal` file?
It is also worth considering to switch to `default-language: GHC2021` because we only support GHC 9.6 and up now.

andreasabel (approved):

exes/Main.hs
andreasabel: `PatternSynonyms` is an essential part of the Haskell language, imo, as it provides a robust syntax for importing constructors (as in `import M (pattern c)`), so I think it should be _always on_.

But maybe this can be taken in another PR.

src/Distribution/Server/Framework/Instances.hs
phadej: Also, removing `HaskellSuite` constructor was completely unnecessary, AFACIT.

There is e.g. `Helium` and being in this (pseudo)enum is the only reference of it.

exes/Main.hs
phadej:
```
    The ‘pattern’ namespace specifier is deprecated.
    Suggested fix:
      Use the ‘ExplicitNamespaces’ extension
```

andreasabel: Thanks @phadej, for the heads up!  
The `data` namespace is being added in 9.14:
> data is enabled by [ExplicitNamespaces](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/explicit_namespaces.html#extension-ExplicitNamespaces) and requires GHC ≥9.14

src/Distribution/Server/Framework/Instances.hs
andreasabel: > Also, removing `HaskellSuite` constructor was completely unnecessary, AFAICT.

Yeah, maybe.

> There is e.g. `Helium` and being in this (pseudo)enum is the only reference of it.

At least `Helium` once existed, in contrast to `HaskellSuite`: https://hackage.haskell.org/package/helium

andreasabel:
@Bodigrim Can this be merged?  GHC 9.14 alpha is out now.

PR #1433, opened 2025-09-27
package: display Uploaded before Revised

juhp:
A small but useful change to package pages:
list the original Upload date before the Revision date

(I often find myself scanning package pages to find when the version was original released (or revised).
Placing the two dates next to each other makes them easier to find.)


juhp:
Thanks!

Issue #1366, opened 2025-03-04
Cache causing invalid hash for 01-index


chreekat:
https://github.com/haskell/ghcup-hs/actions/runs/13656361157/job/38176422379?pr=1238#step:5:80

```
Released cache lock on C:\/Users/runneradmin/AppData/Roaming/stack/pantry/hackage/hackage-security-lock
Verification loop. Errors in order:
  Invalid hash for <repo>/01-index.tar.gz
  Invalid hash for <repo>/01-index.tar.gz
  Invalid hash for <repo>/01-index.tar.gz
  Invalid hash for <repo>/01-index.tar.gz
  Invalid hash for <repo>/01-index.tar.gz
```

avdv:
related to https://github.com/haskell/hackage-security/issues/123

avdv:
As discussed [here](https://discourse.haskell.org/t/hackage-errors-invalid-hash-s-5170-invalidchunkheaders/12935) this issue is caused by inconsistent information being cached for dependent files.

* `timestamp.json` contains the hash for `snapshot.json`
* `snapshot.json` contains the hash for `01-index.tar.gz`

Now, the JSON files have a max-age of 60 seconds, whereas the tarball has one of 300 seconds.

When the files on the origin server are updated the information which is cached by  Fastly / varnish can get out of sync. It could take up to 60 seconds until timestamp and snapshot are consistent again, and it can take up to 300 seconds until snapshot and tarball are consistent again. 

Apparently, there are only 5 retries (I guess there is no exponential back-off, right?) and this can fail pretty frequently (depending on how many updates happen on hackage throughout the day).

This could be solved in different ways... probably the best thing would be to sent purge requests to varnish. WDYT?

Maybe there a better place to discuss this?

gbaz:
On that fix, we can change the uses of `maxAgeMinutes` in this file https://github.com/haskell/hackage-server/blob/8815bc9e27ec25506952f6e41b38d5293d5d31d6/src/Distribution/Server/Features/Core.hs#L666

avdv:
BTW, here is a workflow run which illustrates how annoying this problem is: https://github.com/tweag/rules_haskell/actions/runs/18094346202

Of 52 CI jobs (those that make use of hackage), *26* failed because the cache was in an inconsistent state (invalid hash for `01-index.tar.gz`).

[...]

PR #1435, opened 2025-09-29
Reduce max-age to 1 minute for tarballs

avdv:
This should reduce the chance that the cache (shared, local) is in an invalid state
and verifying the tarballs fails repeatedly leading to a fatal error after 5 retries.

See #1366


avdv:
Thanks for merging @gbaz !

Is there a way to know when this is deployed?!

Issue #710, opened 2018-03-20
Unable to upload a new version of my LGPL licensed project as it is similar to "AllRightsReserved"


DougBurke:
After today's Hackage update, I get

```
% cabal upload dist/swish-0.9.2.1.tar.gz 
Hackage username: NotMyUserName
Hackage password: DefinitelyNotMyPassword
Uploading dist/swish-0.9.2.1.tar.gz...
Error uploading dist/swish-0.9.2.1.tar.gz: http code 400
Error: Invalid package

This server does not accept packages with 'license' field set to e.g.
AllRightsReserved. See https://hackage.haskell.org/upload for more information
about accepted licenses.
```

which I don't understand since

```
% grep -i license swish.cabalLicense:            LGPL
License-file:       LICENSE 
```

and Hackage was perfectly fine with this license on March 8th when I last uploaded a version (https://hackage.haskell.org/package/swish-0.9.2.0).

For reference the cabal file used when attempting the upload is https://bitbucket.org/doug_burke/swish/src/6d1b4d4fad760b64904dff6230a1612464cf9b24/swish.cabal?at=default&fileviewer=file-view-default

and

```
% cabal --version
cabal-install version 1.24.0.2
compiled using version 1.24.2.0 of the Cabal library 
```

gbaz:
I think I figured it out, but we really should catch this error and document it better. In part of the translation to SPDX now, we call the following logic: https://hackage.haskell.org/package/Cabal-2.2.0.0/docs/src/Distribution.License.html#licenseToSPDX

Note that it falls through on LPGL if it does not have a proper version. So we are stricter than before. You need LGPL-3 or LGPL-2 or the like.

(but yes, that error message is not helpful at all!)

gbaz:
cc @phadej -- maybe we can update some docs here too. the upload page currently points to Distribution.License, which doesn't explain that the version fields are mandatory. So we should update the docs there in Cabal? Also maybe create a better place to give recommended licenses and syntax? In the cabal manual? (On the other hand, `cabal init` does the right thing, so this is really an issue with old legacy packages...)

DougBurke:
Thanks for the pointer - I was able to upload after changing `LGPL` to `LGPL-2.1`.

PR #1441, opened 2025-11-06
Change license acceptance criteria

### LINKED ISSUE #1440: License check rejects FOSS license (in SPDX format)
I tried submitting a package with the license set to `AGPL-3.0-only WITH Universal-FOSS-exception-1.0 OR LicenseRef-commercial` and got the error “This server does not accept packages with 'license' field set to e.g. AllRightsReserved.” #710 talks about this message being misleading in some cases, but I figure in this case it actually determined the license is not FOSS for some reason.

I get that it’s not a trivial license[^1], but
1. the `OR` means that only one side needs to apply, so `LicenseRef-commercial` can be ignored, leaving `AGPL-3.0-only WITH Universal-FOSS-exception-1.0`
2. I’m not sure exactly how to handle `WITH` – SPDX [makes it easy to determine if a license is FSF or OSI approved](https://spdx.org/licenses/), but there’s no equivalent for [exceptions](https://spdx.org/licenses/exceptions-index.html) – my inclination would be to accept them by default, and blacklist any problematic ones as you come across them. `Universal-FOSS-exception-1.0` gives additional rights, so I think shouldn’t be blacklisted.

[^1]: I have another package I’m about to publish with the slightly _more_ complex `(AGPL-3.0-only WITH Universal-FOSS-exception-1.0 OR LicenseRef-commercial) AND BSD-3-Clause` license. This one should also pass, IMO, since it can effectively be reduced to `AGPL-3.0-only AND BSD-3-Clause` (with the whitelisting proposal I made above).

sellout:
- allow license exceptions by default – specific ones can be added to a reject list (fixes #1440)
- allow `+` licenses
- report the portion of the license that is unacceptable in the error message (improves #710, but could be more explicit about why licenses are rejected)
- extract the constant allow/reject lists from the evaluation function

### DIFF
--- a/src/Distribution/Server/Packages/Unpack.hs
+++ b/src/Distribution/Server/Packages/Unpack.hs
@@ -51,6 +51,8 @@ import Distribution.Server.Util.ParseSpecVer
 import qualified Distribution.SPDX as SPDX
 import qualified Distribution.SPDX.LicenseId as SPDX.LId
 import qualified Distribution.License as License
+import Distribution.Pretty
+         ( prettyShow )
 
 import Control.Monad.Except
          ( ExceptT, runExceptT, MonadError, throwError )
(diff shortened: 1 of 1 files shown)


sellout:
I’m happy to change any specific parts of this – e.g., change the license exceptions to an accept list instead of a reject list (as long as I get to add `Universal-FOSS-exception-1.0` to the accept list) or implement more specific rejection of `+` or other deprecated licenses.

PR #1444, opened 2025-12-03
Clean up warnings

spencerjanssen:
`hackage-server` builds with lots of warnings. They mostly fall in a few categories:

- Derived `Typeable` instances
- Partial functions like `head` and `tail` which warn in recent versions of GHC and `base`
- Incomplete pattern matches

Is a PR to clean up all the warnings welcome?

### DIFF
--- a/src/Distribution/Server/Features/Core.hs
+++ b/src/Distribution/Server/Features/Core.hs
@@ -763,7 +763,8 @@ coreFeature ServerEnv{serverBlobStore = store} UserFeature{..}
 
     serveCabalFileRevisionName :: DynamicPath -> ServerPartE Response
     serveCabalFileRevisionName dpath = do
-      pkgid1 <- packageTarballInPath dpath
+      -- TODO bug? Maybe something to do with #1439
+      _pkgid1 <- packageTarballInPath dpath
       pkgid2 <- packageInPath dpath
       guard (pkgVersion pkgid2 == pkgVersion pkgid2)
       pkginfo <- packageInPath dpath >>= lookupPackageId


src/Data/StringTable.hs
```diff
@@ -42,7 +42,7 @@ instance MemSize (StringTable id) where
 lookup :: Enum id => StringTable id -> String -> Maybe id
 lookup (StringTable bs tbl) str = binarySearch 0 (topBound-1) (BS.pack str)
   where
-    (0, topBound) = A.bounds tbl
+    (_assumedZero, topBound) = A.bounds tbl
```
ysangkok: I don't think this is an improvement, since the algorithm now runs further if the `_assumedZero` isn't actually zero. But should it run further? It's basically the question of whether you want `assert`s in prod or not.

Personally, I think partiality is better than an incorrect result. Even though the compiler warns for partiality, but it doesn't warn for incorrectness. Of course, it would be even better to prove the algorithm correct. But that isn't always feasible, and the lack of dependent types make it extra difficult.

ysangkok (changes requested):

spencerjanssen:
I should have explained the value I see in addressing these warnings. I find  `-Wall` to be very useful when hacking on Haskell code bases, the warnings often expose real bugs. However, the value of `-Wall` is undermined when there are many existing warnings -- it's too hard to find the newly introduced warnings. hackage-server has something like 60 warnings (excluding `Typeable` deriving) which is way too much noise to be useful.

Having said that, I know from experience that reviewing dozens of small refactors is not pleasant, and perhaps others don't share my outlook regarding warnings. Do we see value in aiming for near-zero warnings?

PR #1449, opened 2026-01-05
Support tar 0.7

### LINKED ISSUE #1430: Support tar-0.7.0.0
Note that this may require more than simply bumping bounds, as there's an API change (sadly, not in the changelog) that avoids running out of file descriptors: see https://github.com/haskell/cabal/pull/11209/files.

spencerjanssen:
Resolves #1430


spencerjanssen:
Fixed the Nix build, CI should pass now.

Addressed the two instances of `Tar.write . Tar.pack` I could find. As far as I can tell, the other instances of `Tar.write` don't have this particular file descriptor exhaustion issue.

PR #1450, opened 2026-01-05
Remove Typeable deriving clauses

spencerjanssen:
Since GHC 7.10 Typeable is derived automatically.


spencerjanssen:
Accidentally deleted a `foldl'` import needed for older GHCs. Fixed, CI should pass now.

PR #1453, opened 2026-01-12
Fix #1422: Fix unreadable markdown tables in dark mode

### LINKED ISSUE #1422: markdown tables from rendered README are unreadable in darkmode
The tables in rendered READMEs on overview sites is unreadable (white on white) with darkmode.

adpad-13:
Previously, tables in rendered READMEs had low contrast (white on white or black on black) in dark mode. This change adds high-contrast colors using the Hackage purple theme to ensure readability.

### DIFF
--- a/datafiles/static/hackage.css
+++ b/datafiles/static/hackage.css
@@ -1314,9 +1314,4 @@ a.deprecated[href]:visited {
       background-color: #5E5184;
       color: #e0e0e0;
   }
-
-  tbody td{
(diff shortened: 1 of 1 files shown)


gbaz:
The tables don't appear unreadable to me? (first screenshot firefox, second chrome)

image

image

gbaz:
Got it, that explains that we would want to fix thead. Is there an example that motivates why we want to change tbody / td?

PR #1493, opened 2026-04-16
Add user agent for hackage-cli

andreasabel:
Like for `stack`, `hackage-cli` also has its own user agent.


andreasabel:
@ysangkok Thanks for merging!  Can you ping me when this commit goes live?

ysangkok:
@andreasabel I can't since I don't control deployments at all. But maybe Gershom can tell you, or you can see when it shows up on the `central-server` branch.

gbaz:
we've redeployed hackage with this merged.

PR #1495, opened 2026-04-22
Remove legacypasswd

isovector:
This PR removes the LegacyPasswd feature, as suggested by https://github.com/haskell/hackage-server/issues/1153#issuecomment-1370308832


gbaz:
This completes a cleanup of the disabling of the feature first done in https://github.com/haskell/hackage-server/pull/1165 I believe.

PR #1498, opened 2026-04-28
Remove PreferredVersions functionality

### LINKED ISSUE #1345: Distinguish between deprecated and non-preferred versions in the package index
Hackage shows deprecated and preferred versions on each package page, but it seems to combine the two types of version ranges in the package index. This means that cabal can't distinguish between versions that are deprecated and versions that are only non-preferred. It would be great to add more functionality to cabal related to identifying and avoiding deprecated versions. Here is an example of an issue that could be solved: https://github.com/haskell/cabal/issues/10097

Adding preferred versions to the index: https://github.com/haskell/hackage-server/blob/5b0d3b059410360651fc706fc7051ec46d5a7970/src/Distribution/Server/Features/PreferredVersions.hs#L344-L350

Formatting deprecated/preferred versions for one package: https://github.com/haskell/hackage-server/blob/5b0d3b059410360651fc706fc7051ec46d5a7970/src/Distribution/Server/Features/PreferredVersions.hs#L411-L413

`PreferredInfo`, with `sumRange` field for merged version ranges: https://github.com/haskell/hackage-server/blob/5b0d3b059410360651fc706fc7051ec46d5a7970/src/Distribution/Server/Features/PreferredVersions/State.hs#L30-L34

Merging version ranges: https://github.com/haskell/hackage-server/blob/5b0d3b059410360651fc706fc7051ec46d5a7970/src/Distribution/Server/Features/PreferredVersions/State.hs#L39-L44

### LINKED ISSUE #1374: Consider removing the preferred versions feature
@phadej and @gbaz suggested removing preferred versions in #1345, so I'm opening a new issue to discuss whether the feature is still needed. Is there a use case for preferring a version that is not the latest, when the latest version shouldn't simply be deprecated?

isovector:
This PR removes the "preferred versions" part of the `PreferredVersions` feature. 
My motivation for cleaning this up is to minimize the work required to accomplish #1486.

Closes https://github.com/haskell/hackage-server/issues/1345
Fixes https://github.com/haskell/hackage-server/issues/1374

After this PR, all of hackage should behave as if there were never any preferred versions. 

I've replaced the cached `sumRange` with an on-demand function in order to invalidate the cache. I haven't benchmarked this, but could do something less stupid here if you think it's important.


src/Distribution/Server/Features/PreferredVersions/Types.hs
```diff
@@ -2,8 +2,4 @@
 
 module Distribution.Server.Features.PreferredVersions.Types where
 
-import Data.SafeCopy (base, deriveSafeCopy)
-
-data VersionStatus = NormalVersion | DeprecatedVersion | UnpreferredVersion deriving (Show, Eq, Ord, Enum)
-
-$(deriveSafeCopy 0 'base ''VersionStatus)
```
isovector, line 9: Instance removed to prove that we aren't serializing this thing anywhere

src/Distribution/Server/Features/PreferredVersions/State.hs
```diff
@@ -22,24 +22,37 @@ data PreferredVersions = PreferredVersions {
...
-consolidateRanges ranges depr =
+
+sumRange :: PreferredInfo -> Maybe VersionRange
+sumRange (PreferredInfo ranges depr _) =
     let range = simplifyVersionRange $ foldr intersectVersionRanges anyVersion (map notThisVersion depr ++ ranges)
     in if isAnyVersion range || isNoVersion range
         then Nothing
         else Just range
 
-data VersionStatus = NormalVersion | DeprecatedVersion | UnpreferredVersion deriving (Show, Eq, Ord, Enum)
```
isovector, line 41: This was a bad merge in #1492; the type got copied into `.Types` but never ended up being used.

PR #1499, opened 2026-04-28
Structured use of PkgInfo

isovector:
This PR refactors all uses of `PkgInfo` so as to not use the internal `acid-state` selectors `pkgMetadataRevisions` and `pkgTarballRevisions`, preferring instead high-level semantic accessors.

The motivation here is be much more precise about exactly what data is required in the existing implementation. In a post `acid-state` world, we are not going to want to fetch the state of the world and filter afterwards.

Implementation note: I put this together by turning on `-XNoFieldSelectors`, such that every use of the selectors became an error. I then traced through each use and rewrite in terms of projections from `Distribution.Server.Packages.Types`.

### DIFF
--- a/src/Distribution/Server/Features/Sitemap.hs
+++ b/src/Distribution/Server/Features/Sitemap.hs
@@ -19,7 +19,6 @@ import Distribution.Text (display)
 import Distribution.Server.Packages.Types
 
 import qualified Distribution.Server.Packages.PackageIndex as PackageIndex
-import qualified Data.Vector as Vec
 import qualified Data.Map  as Map
 import qualified Data.Text as T
 import Data.ByteString.Lazy (ByteString)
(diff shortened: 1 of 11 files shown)


src/Distribution/Server/Features/Sitemap.hs
```diff
@@ -233,7 +232,7 @@ generateSitemap serverBaseURI pageBuildDate alltags pkgIndex docIndex cachedTarI
         [ ( prefixPkgURI ++ display (packageName pkg)
           , uploadtime)
         | pkg <- map head pkgss
-        , let (_, (uploadtime, _user)) = Vec.head (pkgMetadataRevisions pkg)
+        , let (uploadtime, _user) = pkgLatestUploadInfo pkg
```
isovector, line 235: Note that `Vec.head` was the *first* revision; this is now the latest revision. The use of `head` here seemed like a bug to me, but I can keep the old behavior if desired.

PR #1502, opened 2026-05-04
Separate out Package utilities & enforce better typing discipline

isovector:
This PR moves the `Utilities` section out of `Distribution.Server.Packages.Types` and into `Distribution.Server.Packages.Utils`. It then adds:

```diff
+data MetadataRevision = MetadataRevision
+  { metaRevCabalFile :: CabalFileText
+  , metaRevInfo :: UploadInfo
+  }
+  deriving stock Show

-type UploadInfo = (UTCTime, UserId)
+data UploadInfo = UploadInfo
+  { uploadInfoTime :: UTCTime
+  , uploadInfoUser :: UserId
+  }
+  deriving stock Show
```

These new datatypes used to be anonymous tuples, and their new existence gives an easy-to-target API for the work in #1486. This PR is careful to not use the new `MetadataRevision` constructor anywhere, so that it can be trivially swapped out for a structure that can be pulled out of the database.

In addition, we keep `OldUploadInfo` around, since swapping it inside of the `acid-state` would be a breaking serialization change. Furthermore, all of the utility implementations are copied verbatim into `Packages.Utils.Acid`, since they're still required for `*.State` maintenance, and will diverge more dramatically in upcoming PRs.  

This PR also changes the type of `pkgDesc`:

```diff
-pkgDesc :: PkgInfo -> GenericPackageDescription
+pkgDesc :: MetadataRevision -> GenericPackageDescription
```

which makes it clear that it is a function of a *specific revision*, rather than of the package-as-a-whole. This in turn revealed some bugs in the feed code, which was using the wrong dates. I will send a separate PR fixing those issues.

### DIFF
--- a/src/Distribution/Server/Features/PackageFeed.hs
+++ b/src/Distribution/Server/Features/PackageFeed.hs
@@ -102,7 +102,7 @@ renderPackageFeed users hostURI now name pkgs = RSS title uri desc (channel upda
         desc = "New releases of package '" ++ unPackageName name ++ "' on Hackage."
         items = feedItems users uri <$> pkgs
         uri = hostURI { uriPath = "/package/" ++ display name }
-        updated = maybe now (fst . pkgOriginalUploadInfo . fst) (listToMaybe pkgs)
+        updated = maybe now (pkgOriginalUploadTime . fst) (listToMaybe pkgs)
 
 channel :: UTCTime -> [RSS.ChannelElem]
 channel updated =
@@ -132,8 +132,8 @@ feedItems users hostURI (pkgInfo, chlog) =
           , d "Maintainer" $ maintainer pd
           ] +++ XHtml.hr +++ chlog
         pkgName = display (pkgInfoId pkgInfo)
-        (time, uploaderId) = pkgOriginalUploadInfo pkgInfo
+        UploadInfo time uploaderId = pkgOriginalUploadInfo pkgInfo
         timestr = formatTime defaultTimeLocale "%Y-%m-%dT%H:%M:%SZ" time
         uploader = display $ Users.userIdToName users uploaderId
-        pd = packageDescription (pkgDesc pkgInfo)
+        pd = packageDescription $ pkgDesc $ pkgLatestRevision pkgInfo
         d dt dd = XHtml.dterm (XHtml.toHtml dt) +++ XHtml.ddef (XHtml.toHtml dd)
--- a/src/Distribution/Server/Features/PackageFeed.hs
+++ b/src/Distribution/Server/Features/PackageFeed.hs
@@ -8,6 +8,7 @@ import Distribution.Server.Features.Users
 import Distribution.Server.Framework
 import Distribution.Server.Packages.ChangeLog
 import Distribution.Server.Packages.Types
+import Distribution.Server.Packages.Utils
 import qualified Distribution.Server.Users.Users as Users
 import Distribution.Server.Users.Users (Users)
 import Distribution.Server.Util.Parse (unpackUTF8)
--- a/src/Distribution/Server/Features/PackageList.hs
+++ b/src/Distribution/Server/Features/PackageList.hs
@@ -264,7 +264,7 @@ listFeature CoreFeature{..}
         let pkgs = PackageIndex.lookupPackageName index pkgname
         case pkgs of
            [] -> modifyMemState itemCache (Map.delete pkgname)
-           _  -> modifyItem pkgname (updateDescriptionItem $ pkgDesc $ last pkgs)
+           _  -> modifyItem pkgname (updateDescriptionItem $ pkgDesc $ pkgLatestRevision $ last pkgs)
         runHook_ itemUpdate (Set.singleton pkgname)
 
     refreshDownloads = do
@@ -284,7 +284,7 @@ listFeature CoreFeature{..}
     constructItem :: PkgInfo -> IO (PackageName, PackageItem)
     constructItem pkg = do
         let pkgname = packageName pkg
-            desc = pkgDesc pkg
+            desc = pkgDesc $ pkgLatestRevision pkg
         intRevDirectCount <- revDirectCount pkgname
         users <- queryGetUserDb
         tags  <- queryTagsForPackage pkgname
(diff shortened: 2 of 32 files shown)


src/Distribution/Server/Features/PackageFeed.hs
```diff
@@ -101,7 +102,7 @@ renderPackageFeed users hostURI now name pkgs = RSS title uri desc (channel upda
         desc = "New releases of package '" ++ unPackageName name ++ "' on Hackage."
         items = feedItems users uri <$> pkgs
         uri = hostURI { uriPath = "/package/" ++ display name }
-        updated = maybe now (fst . pkgOriginalUploadInfo . fst) (listToMaybe pkgs)
+        updated = maybe now (pkgLatestUploadTime . fst) (listToMaybe pkgs)
```
isovector: It seems wrong that `updated` should be the _original_ time

src/Distribution/Server/Features/PackageFeed.hs
```diff
@@ -131,8 +132,8 @@ feedItems users hostURI (pkgInfo, chlog) =
           , d "Maintainer" $ maintainer pd
           ] +++ XHtml.hr +++ chlog
         pkgName = display (pkgInfoId pkgInfo)
-        (time, uploaderId) = pkgOriginalUploadInfo pkgInfo
+        UploadInfo time uploaderId = pkgLatestUploadInfo pkgInfo
```
isovector: Likewise, it seems incorrect that the feed items should be based on the original upload times. Especially given that the package description was fetched from the latest revision.

src/Distribution/Server/Features/PackageList.hs
```diff
@@ -157,7 +158,7 @@ initListFeature _env = do
         modifyItem pkgname $ \x ->
             updateReferenceVersion prefsinfo allVersions $
               x
-                { itemLastUpload = fst (pkgOriginalUploadInfo pkg)
+                { itemLastUpload = pkgLatestUploadTime pkg
```
isovector: Last upload time seems like it should be the latest time

src/Distribution/Server/Features/PackageList.hs
```diff
@@ -299,7 +300,7 @@ listFeature CoreFeature{..}
           , itemDeprecated = deprs
           , itemDownloads  = cmFind pkgname downs
           , itemVotes      = votes
-          , itemLastUpload = fst (pkgOriginalUploadInfo pkg)
+          , itemLastUpload = pkgLatestUploadTime pkg
```
isovector: Likewise, last uploaded should probably be latest.

src/Distribution/Server/Pages/Recent.hs
```diff
@@ -201,7 +202,7 @@ recentFeed conf users hostURI now pkgs = RSS
     (start,end) = pageIndexRange conf
     desc = "Showing " ++ show start ++ " - " ++ show end ++ " most recent additions to Hackage, the Haskell package database."
     pkgList = paginate conf pkgs
-    updated = maybe now (fst . pkgOriginalUploadInfo) (listToMaybe pkgList)
+    updated = maybe now pkgOriginalUploadTime (listToMaybe pkgList)
```
isovector, line 205: Should probably be latest upload time

src/Distribution/Server/Pages/Recent.hs
```diff
@@ -214,10 +215,10 @@ recentRevisionsFeed conf users hostURI now pkgs = RSS
     (start, end) = pageIndexRange conf
     desc = "Showing " ++ show start ++ " - " ++ show end ++ " most recent revisions to cabal metadata in Hackage, the Haskell package database."
     pkgList = paginate conf pkgs
-    updated = maybe now (fst . pkgOriginalUploadInfo) (listToMaybe pkgList)
+    updated = maybe now pkgOriginalUploadTime (listToMaybe pkgList)
```
isovector, line 218: should be latest

src/Distribution/Server/Pages/Recent.hs
```diff
@@ -239,12 +240,12 @@ releaseItem users hostURI pkgInfo =
   where
     uri   = hostURI { uriPath = packageURL pkgId }
     title = display (packageName pkgId) ++ " " ++ display (packageVersion pkgId)
-    body  = fromShortText $ synopsis (packageDescription (pkgDesc pkgInfo))
+    body  = fromShortText $ synopsis $ packageDescription $ pkgDesc $ pkgOriginalRevision pkgInfo
```
isovector: Uses original description based on using the original upload info, but I'm not sure about this change.

gbaz (approved):

Ericson2314:
Any reason this wasn't merged?

isovector:
@Ericson2314 I don't have the necessary bits, and once I realized the refactoring effort isn't likely to be fruitful, it didn't seem worth pushing on.

Ericson2314:
Well if you consider this done, than I suppose @gbaz or someone can merge it if he wants accordingly.

ysangkok:
@Ericson2314 [This section](https://github.com/tweag/tech-proposals/blob/tweag/proposal-modernising-hackage-server/proposals/0000-modernising-hackage-server.md#incremental-migration-by-tweag) of the tech proposal draft suggests that "incremental refactoring is not feasible". I merged previous PRs before the tech proposal was announced, and I still thought an iterative refactoring was the plan. Now that a more comprehensive plan has been announced, I think the people responsible for the rewrite should have full access if it is to be executed, and it doesn't make sense for me to review iterative work like this, given that I am not intimately familiar with the details of the rewrite, and I won't have time to properly review it. Since I do not have access to deployments, all my previous work has been based on local testing, and then resolving remaining issues based on feedback from Gershom. I don't think that model works with the "live migration" model announced in the tech proposal. Please note that this just explains my own inaction, nothing else. I have no particular opposition to the code of this PR, and I am very grateful that Tweag is contributing. 

Regarding the processing of the proposal: The tech proposal repo dictates that it is HF leadership that has the final say on whether proposals are accepted or not. I don't know if HF leadership has started reviewing the proposal yet, but surely it wouldn't make sense to start executing the draft plan before it has been vetted?

Ericson2314:
I can understand not bothering to finish this PR if a different approach is being gone with. However, it is both non-draft and so appears finished (?) and approved. So it may be just a matter of hitting the button.

I liked the look of it / liked the changes as described in the PR description, so I made my SQL refactor on top of it.

PR #1507, opened 2026-06-06
Redact admin notes in Scrub backup mode

jacobprudhomme:
# Overview
Fixes HSEC-2026-0005. Now scrubs admin notes of sensitive information in Scrub backup mode


jacobprudhomme:
As a first-time contributor, I have a question: should I add tests for this behaviour? If so, where/how would it be best to do so?

PR #1509, opened 2026-06-07
constraints: actually forbid the unwanted packages

frasertweedale:
Intent of "forbidden packages" constraint in `cabal.project` is not effected.  The current constraint does not prevent inclusion in build plan - it only ensures the packages are build without the `none` flag.

Change to a `< 0` version constraint to actually prevent the unwanted packages from being used.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -9,8 +9,8 @@ packages:
 -----------------------------------------------------------------------------
 -- Anti-constraints
 
--- make sure we don't accidentally pick up blacklisted packages
-constraints: foundation -none, cryptonite -none, cryptohash -none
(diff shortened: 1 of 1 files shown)


philderbeast:
This is the right thing to do. Even though project files are not versioned, cabal files are and the `-none` syntax is deprecated and should be replaced with `<0`:

> Remove -any and -none syntax for version ranges Use >=0 and <0 respectively.
> -- [cabal-version: 3.4](https://cabal.readthedocs.io/en/latest/file-format-changelog.html#cabal-version-3-4)

https://github.com/haskell/hackage-server/blob/8b5f7712184bce347980fe1ddda1708328064dd6/hackage-server.cabal#L1-L2

PR #1510, opened 2026-06-07
convert to crypton

frasertweedale:
*crypton* is today the dominant cryptographic primitives "kitchen
sink" library.  There is a need and want of additional cryptographic
primitives in *hackage-server*, including HMAC (various use cases)
and, later, additional public key signature algorithms to support
OpenID Connect (for trusted publishing) and Passkeys/webauthn login.

This change migrates *hackage-server* to *crypton*.  The changes are
kept to a minimum without any refactoring.

Dependencies on *cryptohash-md5* and *cryptohash-sha256* have been
dropped.  It would have been nice to drop *ed25519*, but without it
we would be depending on internal implementation details to define
the `MemSize` instance for the *hackage-security* `Key` type.

**NOTE: not tested yet.  Made PR for early review and for others to
start building upon.**

### DIFF
--- a/hackage-server.cabal
+++ b/hackage-server.cabal
@@ -431,7 +431,7 @@ library
     , HStringTemplate       ^>= 0.8
     , HTTP                   >= 4000.3.16 && < 4000.6
     , http-client           ^>= 0.7       && < 0.8
-    , http-client-tls       ^>= 0.3
+    , http-client-tls       ^>= 0.4
     , http-types             >= 0.10      && < 0.13
     , QuickCheck             >= 2.14      && < 2.19
     , acid-state            ^>= 0.16
@@ -451,10 +451,8 @@ library
         -- commonmark-0.2 needed by commonmark-extensions-0.2.2
     , commonmark-extensions ^>= 0.2.2
         -- Note: 0.2.2 added footnoteSpec to gfmExtensions
-    , cryptohash-md5        ^>= 0.11.100
-    , cryptohash-sha256     ^>= 0.11.100
+    , crypton                >= 1.1.4 && < 1.2
     , csv                   ^>= 0.1
-    , ed25519               ^>= 0.0.5
     , hackage-security       >= 0.6 && < 0.7
         -- N.B: hackage-security-0.6.2 uses Cabal-syntax-3.8.1.0
         -- see https://github.com/haskell/hackage-server/issues/1130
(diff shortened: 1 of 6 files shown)


blackheaven (approved):

gbaz:
memsize is i believe an approximate heuristic and its fine to give an
instance thats a crude estimate

frasertweedale:
> memsize is i believe an approximate heuristic and its fine to give an instance thats a crude estimate

@gbaz ~~I'd prefer to leave it alone. The `ed25519` package doesn't interfere with anything else, and we only use the data constructors.  If you insist, I will replace the existing calculations with a "magic number" and try not to lose sleep over it  :sweat_smile:~~

Never mind all that.  I see that it is used for memory usage monitoring and am satisfied that estimates are indeed fine.  I removed the dependency and replaced it with a magic number.

ysangkok:
seems like this is blocked on

* https://github.com/well-typed/cborg/issues/379

it seems like it isn't an issue on master since it was building with tls-1.9.0: https://github.com/haskell/hackage-server/actions/runs/27108585879/job/80002069176 , which does not need `serialise`

hackage-server.cabal
```diff
@@ -451,10 +451,8 @@ library
         -- commonmark-0.2 needed by commonmark-extensions-0.2.2
     , commonmark-extensions ^>= 0.2.2
         -- Note: 0.2.2 added footnoteSpec to gfmExtensions
-    , cryptohash-md5        ^>= 0.11.100
-    , cryptohash-sha256     ^>= 0.11.100
+    , crypton                >= 1.1.4 && < 1.2
```
phadej, line 454: I don't think this is an improvement, `cryptohash-md5` and `cryptohash-sha256` are super simple and stable packages doing one thing well, i.e. providing corresponding hash primitive. `crypton` OTOH is a kitchen sink.

`hackage-security` already depends on `ed25519` and `cryptohash-sha256` so this patch doesn't remove dependencies, it adds them.

If there is need for more crypto stuff, it can be added. I don't think this improves the existing stuff though.

hackage-server.cabal
```diff
@@ -469,6 +467,7 @@ library
     , hslogger              ^>= 1.3.1
     , lifted-base           ^>= 0.2.1
     , mime-mail             ^>= 0.5
+    , ram                    >= 0.19
```
phadej, line 470: No upper bound.

phadej (changes requested):

PR #1514, opened 2026-07-14
Fix for mirroring hackage

isovector:
When attempting to mirror hackage, I received the following error:

```
mirroring OGL-0.0.2
packAscii: only ASCII inputs are supported, but got Administratörer
CallStack (from HasCallStack):
  error, called at ./Codec/Archive/Tar/PackAscii.hs:44:17 in tar-0.7.0.0-KJyftX2xMxi7opmJZqZodj-tar-internal:Codec.Archive.Tar.PackAscii
  packAscii, called at ./Codec/Archive/Tar/Write.hs:188:28 in tar-0.7.0.0-KJyftX2xMxi7opmJZqZodj-tar-internal:Codec.Archive.Tar.Write
```

which then crashed `hackage-mirror`. This PR adds a little `handle` wrapper around the offending call.

### DIFF
--- a/exes/MirrorClient.hs
+++ b/exes/MirrorClient.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+
 module Main (main) where
 
 -- stdlib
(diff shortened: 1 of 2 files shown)


Bodigrim:
@isovector I think I just fixed the issue upstream, in [`tar-0.7.2.0`](https://hackage.haskell.org/package/tar-0.7.2.0). Could you give it a try?

isovector:
@Bodigrim I can confirm that `tar-0.7.2.0` fixes the issue. Thanks!

PR #1516, opened 2026-08-12
Add socket activation support

Ericson2314:
Socket activation is generally nice to have / considered good practice, but this will especially be useful later because it makes parallel local (i.e. via `cabal`, not just inside a Nix build) integration testing easier.

We try socket activation by default, falling back to binding a port ourselves. `--socket-activation-only` and `--no-socket-activation` force either side; `--ip`/`--port` imply the latter, since asking for an address to listen on is a request to bind it ourselves. (Use `--base-uri` if you only meant to change the advertised URI.)

A key implementation detail: `setNonBlockIfNeeded` must be called on the inherited fd so that GHC's IO manager (epoll) can handle `accept`/`recv` correctly in the threaded runtime. Without this, the server starts but connections hang — discovered and fixed via the NixOS VM integration test.

### DIFF
--- a/exes/Main.hs
+++ b/exes/Main.hs
@@ -6,7 +6,7 @@
 module Main where
 
 import qualified Distribution.Server as Server
-import Distribution.Server (ListenOn(..), ServerConfig(..), Server)
+import Distribution.Server (InfallibleListenOn(..), ListenOn(..), ServerConfig(..), Server)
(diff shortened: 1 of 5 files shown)


Ericson2314:
OK yay the CI error was spurious, and this is passing now.

PR #1518, opened 2026-08-24
Batch download counts

### LINKED ISSUE #1342: Download count feature needs optimizing
We keep a record for each download instead of summarizing. On the live instance this counts feature is especially slow to persist and initialize -- we should improve it to reduce memory usage and time spent in restarts.

LaurentRDC:
This patch batches download counts, rather than incurring an acid-state transaction for every download.

This reduces the number of events related to download counts, while creating the potential for some downloads being dropped due to the in-memory cache being flushed only periodically.

Benchmarks included in this patch show much better replay times, and a smaller acid-state event log, proportional to the batching factor. The cost is slightly higher memory usage due to the in-memory download counts cache.

In order to properly shutdown the download counts feature, a "pre-shutdown" step has been added to features, such that download counts can be flushed to disk on graceful shutdown.

This batching does not require anything special when it comes to deployment, as the previous download counts can be modeled as batches of 1. This patch therefore doesn't require a migration.

Partially fixes #1342. The way downloads are stored in per package-day counts forever could also be optimized, but this will require a change in the event log representation.

## Benchmarks

The `DownloadCountBench` benchmarks below distributed 20k downloads over 300 packages. In the `batched` mode, downloads are batched in batches of 3000. Using `individual` mode uses the legacy patch of incrementing the on-disk download count by 1.

Legacy behavior

```console
> cabal run DownloadCountBench -- individual 200000 200
individual (one event per download): 200000 downloads over 200 distinct package versions
  time to record            6.365 s
  event log on disk         34694 KB
  start-up replay           0.388 s
  allocated                 5324 MB
  peak live heap            158 KB
  InMemStats memSize        1122 KB
```

New behavior, where batches of 3000 download counts are flushed at once

```console
> cabal run DownloadCountBench -- batched 200000 200 3000
individual (one event per download): 200000 downloads over 200 distinct package versions
  time to record            0.068 s
  event log on disk         1187 KB
  start-up replay           0.015 s
  allocated                 521 MB
  peak live heap            186 KB
  InMemStats memSize        1122 KB
```

This is an extreme example, but it still shows major performance improvements, roughly proportional to batching.


src/Distribution/Server/Features/DownloadCount.hs
```diff
@@ -164,14 +180,51 @@ downloadFeature CoreFeature{}
     totalPackageDownloads :: MonadIO m => m TotalDownloads
     totalPackageDownloads = readMemState totalDownloadsCache
 
-    registerDownloads = forever $ do
-        pkg    <- readChan downloadStream
+    flushInterval :: Int
+    flushInterval = 60 * 1_000_000 -- 60 seconds
```
LaurentRDC: This is a configuration option. There might be a better place for this

LaurentRDC: I think I'll make this a part of `RunFlags`

ysangkok (changes requested):
I think `Main.hs` is missing the `confDownloadFlush = flushDownloads` line around 349 to actually apply the parsed flag to the configuration.

Also, the flag is declared `noArg` so a user can't actually set the flag...

src/Distribution/Server/Features/DownloadCount.hs
```diff
@@ -125,24 +138,27 @@ downloadFeature :: CoreFeature
...
+                pendingDownloads
+                flushLock
   = DownloadFeature{..}
   where
     downloadFeatureInterface = (emptyHackageFeature "download") {
         featureResources = [ topDownloads downloadResource
                            , downloadCSV
                            ]
-      , featurePostInit  = void $ forkIO registerDownloads
+      , featurePostInit  = void $ forkIO (flushDownloadsLoop serverFlushDownloadsPeriod)
```
ysangkok, line 160: This runs also when restoring from a backup. But it shouldn't. So it would be better to move this `Server.run` to make sure it's only run as part of the normal HTTP serving operation mode.

ysangkok (changes requested):

src/Distribution/Server/Features/DownloadCount.hs
LaurentRDC: I don't understand what's the difference between "restoring from a backup" and normal operations. Do you mean that this flush loop is running during the boot sequence, so-to-speak?

I see a comment in `Distribution.Server.Features` that says that `featurePostInit` runs in backup/restore mode, but it's not clear to me why this is bad per se, given that there already was a loop to process downloads as a post-init hook

LaurentRDC:
> I think `Main.hs` is missing the `confDownloadFlush = flushDownloads` line around 349 to actually apply the parsed flag to the configuration.
> 
> Also, the flag is declared `noArg` so a user can't actually set the flag...

Thanks, indeed I missed this! Addressed in the latest commit

src/Distribution/Server/Features/DownloadCount.hs
ysangkok: Restoring from a backup is a separate command launched with e.g. `hackage-server restore backup.tar`. The difference to 'normal' (i.e. HTTP) operations, is that no server is run, and the archive is just read and ingested into the acid-state database.

If downloads are processed while a backup is getting restored, then that is an issue. But we shouldn't keep doing something bad just because it is already happening.

PR #1519, opened 2026-08-24
Add hs-opentelemetry

isovector:
This PR adds an `--enable-opentelemetry` flag which, when enabled, turns on telemetry around hackage-server's router.

The purpose here is to start collecting traces and get real data on which routes get used in production, and how expensive each is.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -178,6 +178,7 @@ jobs:
           touch cabal.project.local
           echo "packages: ${PKGDIR_hackage_server}" >> cabal.project
           cat >> cabal.project <<EOF
+          allow-newer: lens-family:containers, lens-family-core:containers
           EOF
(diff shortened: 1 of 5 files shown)


phadej:
while telemetry could be useful, I think that `hs-opentelemetry-sdk` is a liability (https://github.com/iand675/hs-opentelemetry/issues/315 - no bounds), especially for a project with effectively zero active maintainers.

ysangkok:
I think logging/monitoring would be great to have, even if the project is under-maintained. But since I do not work on deployments, I will defer to @gbaz on whether this is the right direction.

isovector:
@phadej: very reasonable point. Would hiding it behind a default-off flag assuage some fears here? The reason I ask is that I'm in the middle of chicken-and-egg situation here; I'm trying to get enough data in order to determine if hackage-server-v3 can meaningfully move the needle. Which if it does, is a good argument to continue moving forwards on it. 

Maybe we could gather this information from logs on the existing load balancer / reverse proxy? Perhaps @gbaz would know?

isovector:
Closing due to lack of caring.

PR #1522, opened 2026-09-19
Switch to cabal haddock --haddock-for-hackage

zlonast:


ysangkok:
Interesting. This looks nice as it is a simplification, but I'd also like to hear whether @gbaz think it is risky.

